Issue #200, opened 2021-06-25
Fail to build on Pi4


Exarkuniv:
not sure if it can or not. 
im running Debian-buster 
i have cmake libsdl1.2-dev cmake-curses-gui g++ liballegro5-dev libxft-dev installed

i have used 
git clone to get your software and i did 
mkdir build
  cd build
  
 cmake ..
 
 make 

and i get 
[  4%] Building CXX object CMakeFiles/d2tm.dir/cGame_draw.cpp.o
In file included from /home/pi/RetroPie-Setup/tmp/build/dmaker/include/d2tmh.h:14,
                 from /home/pi/RetroPie-Setup/tmp/build/dmaker/ai.cpp:12:
/home/pi/RetroPie-Setup/tmp/build/dmaker/include/allegroh.h:30:10: fatal error: alfont.h: No such file or directory
 #include <alfont.h>
          ^~~~~~~~~~
compilation terminated.
In file included from /home/pi/RetroPie-Setup/tmp/build/dmaker/include/d2tmh.h:14,
                 from /home/pi/RetroPie-Setup/tmp/build/dmaker/cGame_draw.cpp:1 :
/home/pi/RetroPie-Setup/tmp/build/dmaker/include/allegroh.h:30:10: fatal error: alfont.h: No such file or directory
 #include <alfont.h>
          ^~~~~~~~~~
make[2]: *** [CMakeFiles/d2tm.dir/build.make:63: CMakeFiles/d2tm.dir/ai.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
compilation terminated.
make[2]: *** [CMakeFiles/d2tm.dir/build.make:89: CMakeFiles/d2tm.dir/cGame_draw.cpp.o] Error 1
In file included from /home/pi/RetroPie-Setup/tmp/build/dmaker/building/../include/d2tmh.h:14,
                 from /home/pi/RetroPie-Setup/tmp/build/dmaker/building/cItemBuilder.cpp:1:
/home/pi/RetroPie-Setup/tmp/build/dmaker/building/../include/allegroh.h:30:10: fatal error: alfont.h: No such file or directory
 #include <alfont.h>
          ^~~~~~~~~~
compilation terminated.
make[2]: *** [CMakeFiles/d2tm.dir/build.make:76: CMakeFiles/d2tm.dir/building/cItemBuilder.cpp.o] Error 1
[...]

Issue #58, opened 2020-07-17
Support compilation on Linux


stefanhendriks:
This commit 2e726e9 seemed to introduced it. Try to cherry-pick again.

stefanhendriks:
perhaps mac OS as well? (X86)

SantjagoCorkez:
Even prerequisites fail to build on Debian Buster.

```
Dune-II---The-Maker/dependencies/src/fblend0.5/fblend$ make
gcc -c src/cadd.c -o obj/unix/release/cadd.o -O2 -ffast-math -fomit-frame-pointer -I. -Isrc -Iinclude
gcc -c src/mem_test.c -o obj/unix/release/mem_test.o -O2 -ffast-math -fomit-frame-pointer -I. -Isrc -Iinclude
gcc -c src/csub.c -o obj/unix/release/csub.o -O2 -ffast-math -fomit-frame-pointer -I. -Isrc -Iinclude
gcc -c src/ctrans.c -o obj/unix/release/ctrans.o -O2 -ffast-math -fomit-frame-pointer -I. -Isrc -Iinclude
gcc -c src/radd.c -o obj/unix/release/radd.o -O2 -ffast-math -fomit-frame-pointer -I. -Isrc -Iinclude
gcc -c src/rtrans.c -o obj/unix/release/rtrans.o -O2 -ffast-math -fomit-frame-pointer -I. -Isrc -Iinclude
gcc -c src/2xstretch.c -o obj/unix/release/2xstretch.o -O2 -ffast-math -fomit-frame-pointer -I. -Isrc -Iinclude
gcc -c src/fade.c -o obj/unix/release/fade.o -O2 -ffast-math -fomit-frame-pointer -I. -Isrc -Iinclude
make: *** No rule to make target '/asmdef.c', needed by 'obj/unix/asmdef.exe'.  Stop.
```

BTW, `fblend` uses ASM optimizations. For that I'm not sure if it's OK to build it on amd64 platform. If you're not sure as well, will it be then better to disable that dependency on amd64?

stefanhendriks:
I want to get rid of fblend anyway. It might suffer on the performance side of things, but I want to decouple from Allegro slowly.

I believe fblend routines can be replaced by allegro equivalent calls for now.

mareker80:
On allegro page they claim it works on Linux. Interesting fblend does not compile. Also asmdef.exe? Does not sound like Linux target. Maybe cmake was not run properly?
[...]

PR #369, opened 2021-12-17
Build: introduce CMake import targets for direct dependencies

pbomta:
Hi Stefan, what a cool project! I loved Dune 2 a lot when I was a kid.

You asked for a different approach to link the dependencies in the CMakeLists.txt file 
We can use "import targets" for that. They turn an include directory and a binary artifact into a CMake target that can be added as a dependency to the main executable.

I organized the dependencies too so that each has an "include" directory and a "lib" directory.


stefanhendriks:
This relates to (I think) https://github.com/stefanhendriks/Dune-II---The-Maker/issues/58, possibly affects https://github.com/stefanhendriks/Dune-II---The-Maker/issues/200 

If there is any other issue you mean @pbomta , please let me know, so I can tie this in.

stefanhendriks:
I pulled your PR and updated the `readme.md`. I do wonder if you had to do anything specific with header files to have d2tm compile for you?

stefanhendriks:
I tested on my environment, with updated `cmake`, also removing all my allegro / dependency related files which I had in my `minGW` dir and it compiles flawlessly. It also compiles under command-line (I usually test under CLion, but I really would like to have it IDE-independent obviously). 

Looks like it works fine.

stefanhendriks:
Testing a bit, I see this now (something with alfont). I've seen this before, I'll check the dependencies a bit to see how they are different from my local machine (when i switch back to `master` it works fine):

image

stefanhendriks:
I found a different .lib file to link against, with that the weird artifacts we're gone.

pbomta:
> This relates to (I think) #58, possibly affects #200
> 
> If there is any other issue you mean @pbomta , please let me know, so I can tie this in.
Ah, but I didn't compile on Linux. I have MSYS2 installed on my Windows laptop and then used mingw32 to make use of the binaries that got checked in source control. Good to read that you want to move away from that, by the way.

I went through all the open issues and didn't find any that refers to this fix.

pbomta:
> I do wonder if you had to do anything specific with header files to have d2tm compile for you?

Nope, That was what I wanted to achieve. ;) I didn't want to copy any allegro stuff into my mingw installation. So I could just pull the project from github, run cmake and ninja (I prefer ninja over make) to build it.

PR #372, opened 2021-12-18
fix: enables warnings as errors and -Wswitch

pbomta:
We can enable additional checks by the compiler and not have the same errors creep back in by enabling more warnings and setting warnings as errors. I tried enabling -Wall (there are even more with -Wextra and -Wpedantic), but the amount of warnings were too great to fix in one PR. So I picked the one that warns for missing switch cases, which seemed useful given the C-style of most of the code. In almost all places, fallthrough was perfectly fine, so adding a default solved that. For the switches with only one case, I just made an if statement out of it.

It did find a bug in the logger, where two components were not printed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,6 +82,12 @@ target_link_libraries(${PROJECT_NAME}
     -static-libstdc++
 )
 
+target_compile_options(${PROJECT_NAME}
+  PRIVATE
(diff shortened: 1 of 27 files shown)


stefanhendriks (approved):

stefanhendriks:
looks good to me!

PR #375, opened 2021-12-21
fix: remove unused variables

pbomta:
Next round of additional warnings.
Added some constness to a few variables in large functions too.


stefanhendriks:
lgtm!

PR #377, opened 2021-12-22
build: include what is used in cItemBuilder

pbomta:
We need to get rid of the "uber" header file, IMHO. If we include what is used, then incremental builds are faster as not the entire program needs to be rebuilt.
- Moved some code to .cpp files to hide it from other components. This to reduce the number of includes from header files.
- Introduced forward declarations for private pieces of components that are not needed to interpret the header file. This is also to reduce the number of includes that are needed in header files.
- Found a missing header guard. I put #pragma once there. This is non-standard C++, but all available compilers support it.
Stopped after handling the first .cpp file in the project. Otherwise, the PR would be way too big.

### DIFF
--- a/cGame.h
+++ b/cGame.h
@@ -16,13 +16,13 @@
 #include <observers/cScenarioObserver.h>
 #include <data/cAllegroDataRepository.h>
 
-// forward declaration :/ sigh should really look into this
-class cRectangle;
 class cAbstractMentat;
-class cPlayer;
 class cGameControlsContext;
-class cInteractionManager;
 class cGameState;
+class cInteractionManager;
+class cPlayer;
+class cRectangle;
+class cSoundPlayer;
 
 class cGame : public cScenarioObserver, cMouseObserver {
 
--- a/cGame.h
+++ b/cGame.h
@@ -13,16 +13,16 @@
 #define D2TM_GAME_H
 
 #include <controls/cMouse.h>
-#include <observers/cScenarioObserver.h>
 #include <data/cAllegroDataRepository.h>
+#include <observers/cScenarioObserver.h>
+#include <utils/cRectangle.h>
+#include <utils/cSoundPlayer.h>
 
 class cAbstractMentat;
 class cGameControlsContext;
 class cGameState;
 class cInteractionManager;
 class cPlayer;
-class cRectangle;
-class cSoundPlayer;
 
 class cGame : public cScenarioObserver, cMouseObserver {
 
--- a/include/d2tmc.h
+++ b/include/d2tmc.h
@@ -18,6 +18,34 @@
 
 #ifndef D2TMC_H
 
+#include "cGame.h"
+#include "definitions.h"
+#include "map/cMap.h"
+#include "structs.h"
+#include "utils/cStructureUtils.h"
+
+#include <allegro/palette.h>
+
+class cAbstractStructure;
+class cAllegroDrawer;
+class cBullet;
+class cDrawManager;
+class cMapCamera;
+class cMapEditor;
+class cRandomMapGenerator;
+class cUnit;
+class cParticle;
+class cPlayer;
+class cRegion;
+class cTimeManager;
+
+struct sReinforcement;
+
+struct ALFONT_FONT;
+struct ALMP3_MP3;
+struct BITMAP;
+struct DATAFILE;
+
 // Process 'extern' stuff, so we can access our classes
 extern bool		bDoDebug;
 extern int		iRest;	// rest value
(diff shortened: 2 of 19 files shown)


stefanhendriks:
I agree about the general approach. Its just I never get to it and I also run into some issues where you have cyclic dependencies and I am unsure how to solve that (as in, what the best practice is).

So if you have any ideas about that I am all ears. 

About #pragma once, see my Discord comment: Its fine to use, but then we should do it everywhere so it is consistent. What is the industry standard these days?

cGame.h
```diff
@@ -16,13 +16,13 @@
 #include <observers/cScenarioObserver.h>
 #include <data/cAllegroDataRepository.h>
 
-// forward declaration :/ sigh should really look into this
```
stefanhendriks, line 19: haha love how you removed the comment. In hobby projects I keep my thoughts unfiltered in code haha.

include/d2tmc.h
```diff
@@ -18,6 +18,34 @@
...
 
+#include "cGame.h"
+#include "definitions.h"
+#include "map/cMap.h"
+#include "structs.h"
+#include "utils/cStructureUtils.h"
+
+#include <allegro/palette.h>
+
+class cAbstractStructure;
```
stefanhendriks, line 29: this makes my eyes bleed, but this is how it should be then?

map/cMap.cpp
```diff
@@ -1255,6 +1255,64 @@ cAbstractStructure *cMap::findClosestStructureType(int cell, int structureType,
     return nullptr;
 }
 
+void cMap::cellTakeDamage(int cellNr, int damage) {
```
stefanhendriks, line 1258: nice, good to see them moved into the map class

player/cPlayer.cpp
```diff
@@ -365,6 +365,12 @@ bool cPlayer::hasRadarAndEnoughPower() const {
     return getAmountOfStructuresForType(RADAR) > 0 && bEnoughPower();
 }
 
+std::string cPlayer::asString() const {
+    char msg[512];
+    sprintf(msg, "Player [id=%d, human=%b, sidebar=%d]", this->id, this->m_Human, this->sidebar);
```
stefanhendriks, line 370: does it make sense to render `this->sidebar` as `%d` ?

stefanhendriks:
```
Introduced forward declarations for private pieces of components that are not needed to interpret the header file. This is also to reduce the number of includes that are needed in header files.
```
I am really interested what the best practices are about this. Currently it feels awkward for me to forward declare stuff, but I can't see a way around it sometimes.

include/d2tmc.h
pbomta: I know, right? Thing is, this file declares all the globals. If we do not use forward declarations here, then this will be the header that includes everything. 
I wanted to get rid of d2tmh.h first. Then we can tackle this file. That would require some architectural changes, where, for instance, the *Info structs go into a game or session class that keeps all game data. An object of that type could then be created when a game is started. That would also help with save-files, because if that class has all the necessary data, saving is just a serialization of the contents to disk. But I digress. :)

Baby steps, I'm afraid.

player/cPlayer.cpp
pbomta: No! That's actually a bug. I'll fix it, as the code is being touched in this PR.

include/d2tmc.h
stefanhendriks: cool, since I really would like saving at some point. Although atm I was thinking of having a separate data structure for that with all kinds of mapping in between. (to keep savefiles sane)

pbomta:
> I am really interested what the best practices are about this. Currently it feels awkward for me to forward declare stuff, but I can't see a way around it sometimes.

The C++ Core Guidelines say nothing about forward declarations. The Google Style guide recommends against them. Mainly because if a class changes, the forward declarations may become redundant and there is no tool that checks this. I have read evidence that it speeds up the build (https://gieseanw.wordpress.com/2018/02/25/the-joys-of-forward-declarations-results-from-the-real-world/ for instance).

But it's a bit of an advanced technique, in the sense that the programmer needs to know what they can forward declare and they must also remember to clean them up while refactoring. Header files are less likely to be forgotten.

Thinking about it again, perhaps we should not use forward declarations, because we want to make this repository friendly for contributors. I would like to make an exception for that large header for the globals, though. At least, as long as that file still exists. :)

pbomta:
> About #pragma once, see my Discord comment: Its fine to use, but then we should do it everywhere so it is consistent. What is the industry standard these days?

Every project does it differently, as far as I know. I'll replace the pragma once with header guards for this PR for consistency. We can then later switch.

stefanhendriks:
@pbomta I ran into an issue with duplicate includes, (in my own PR), and I decided to add a `#pragma once` there, since it was for a structure (`cConstYard`), I did them all at once (so all structures have `pragma once` now).

Issue #367, opened 2021-12-08
Redo mouse/keyboard logic


stefanhendriks:
(originally: Holding CTRL and pressing somewhere for rally-point simply no longer works.)

Changes for #290 broke setting rally-point behavior. While trying to fix that it became obvious that event-driven vs 'everything in one big mess' driven bite each other. In other words, it is best to redo this using the event-driven way.

## TODO
- [x] select structure
- [x] select structure while having a group of units selected
- [x] deselect structure (won't do)
- [x] select / deselect unit (single click)
- [x] select multiple units (box-selecting) from "normal" mouse state
- [x] select multiple units (box-selecting) from "units already selected" mouse state
- [x] repair state
   - [x] repair unit (press `R` to switch to that state, press `R` again, or right-mouse click to leave Repair state)
   - [x] repair structure 
- [x] add units to selection when holding `SHIFT` 
- [x] place structure (mouse state)
- [x] deploy super weapon (mouse state) (deferred to #399)
- [x] force attack (hold key: `CTRL`)
- [x] force move (?)
- [x] set rally-point (with right-mouse button instead of CTRL-click)
- [x] react to keypress (key observer, like mouse)?
   - [x] create groups (`CTRL-1/5`)
   - [x] add to group (hold `SHIFT`)
   - [x] select units by group nr
   - [x] (required: report all keys pressed, or held for event, so that we can know if CTRL-1 is pressed for instance))

PR #379, opened 2021-12-22
Refactor: Untangle the move and click logic

stefanhendriks:
This is for #367 

The gist is that draw/game logic is still tied with these pieces of code, and as we're migrating towards an event based approach things break due two the 2 conflicting approaches.

So basically the only solution is to make sure this is working with the event based model.


stefanhendriks:
Ok, so basically I am trying to rewrite this a bit so it makes sense. The draw/input/logic codepaths are so tangled up that this is hard to get right.

So, this needs a bit of rethinking.

While we're at it, might as well think about input vs mouse action abstraction. Ie command pattern:

https://gameprogrammingpatterns.com/command.html

controls/mousestates/cMouseState.h
```diff
@@ -0,0 +1,32 @@
+#ifndef D2TM_CMOUSESTATE_H
+#define D2TM_CMOUSESTATE_H
+
+#include "observers/cInputObserver.h"
+#include "controls/cMouse.h"
+//#include "controls/cGameControlsContext.h"
+
+// forward declaration
+class cGameControlsContext;
+class cPlayer;
```
pbomta, line 10: Yes, cPlayer is hard to include. I guess for the time being, we'll have to forward declare the class everywhere.

controls/mousestates/cMouseState.h
```diff
@@ -0,0 +1,32 @@
+#ifndef D2TM_CMOUSESTATE_H
+#define D2TM_CMOUSESTATE_H
+
+#include "observers/cInputObserver.h"
+#include "controls/cMouse.h"
+//#include "controls/cGameControlsContext.h"
```
pbomta, line 6: Including it should work, as long as we keep forward declaring cPlayer.

controls/mousestates/cMouseState.h
```diff
@@ -0,0 +1,32 @@
...
+
+// forward declaration
+class cGameControlsContext;
+class cPlayer;
+
+class cMouseState : public cInputObserver {
+
+public:
+    cMouseState(cPlayer * player, cGameControlsContext *context, cMouse * mouse);
+    ~cMouseState();
```
pbomta, line 16: This destructor can be left out completely. The compiler will generate the correct default.

observers/cInputObserver.h
```diff
@@ -0,0 +1,17 @@
...
+#include "cMouseObserver.h"
+#include "../include/sKeyboardEvent.h"
+
+class cInputObserver : public cMouseObserver {
+
+public:
+    virtual ~cInputObserver() = default;
+
+    virtual void onNotifyMouseEvent(const s_MouseEvent &event) override = 0;
+    virtual void onNotifyKeyboardEvent(const s_KeyboardEvent &event) = 0;
```
pbomta, line 13: This class hierarchy is a little confusing now. I would suggest creating a separate mouse observer and keyboard observer.
For convenience, one could derive an input observer from both, but we could also derive concrete implementations from both. Fortunately, C++ supports multiple inheritance. :)

include/enums.h
```diff
@@ -258,6 +258,13 @@ enum eMouseEventType {
 	MOUSE_SCROLLED_DOWN
 };
 
+enum eKeyboardEventType {
```
pbomta, line 261: I would suggest enum classes for everything. This increases type safety and eliminates name clases. The values no longer have to be prepended with the enum name. So in this case, that would be:
`enum class eKeyboardEventType {
    None,
    Hold,
    Pressed
};`
Usage is then: `eKeyboardEventType::None`.

include/sKeyboardEvent.h
```diff
@@ -0,0 +1,37 @@
+#ifndef D2TM_SKEYBOARDEVENT_H
+#define D2TM_SKEYBOARDEVENT_H
+
+#include "allegroh.h"
+
+#include <string>
+#include "../utils/cPoint.h"
```
pbomta, line 7:
```suggestion
#include "utils/cPoint.h"
```
CMake will take care of the path. And this file can be moved around to a different directory depth and things will still build.

cGame_logic.cpp
```diff
@@ -1034,177 +987,171 @@ bool cGame::setupGame() {
...
     m_dataRepository = new cAllegroDataRepository();
     allegroDrawer = new cAllegroDrawer(m_dataRepository);
 
     // randomize timer
-	unsigned int t = (unsigned int) time(0);
-	char seedtxt[80];
-	sprintf(seedtxt, "Seed is %u", t);
-	logbook(seedtxt);
-	srand(t);
+    unsigned int t = (unsigned int) time(0);
```
pbomta, line 1102: Nit: in modern C++, this is:
```suggestion
    auto t = static_cast<unsigned int>(time(0));
```
Mentions the type only once and on the right side (as usual). There are whole talks about this. :) The style is called: Almost Always Auto.

cGame_logic.cpp
```diff
@@ -1698,18 +1667,94 @@ void cGame::setNextStateToTransitionTo(int newState) {
...
+                                game.getColorFadeSelected(255, 255, 255));
     }
 
     mouse->draw();
 }
 
-void cGame::onCombatMouseEventMovedTo(const s_MouseEvent &event) {
-    // not needed?
+void cGame::takeScreenshot() {
+    char filename[25];
```
pbomta, line 1685: Ideally, we'll use the std::string type, as that has no upper bound. But I remember an issue where lib fmt is considered. That's probably a better idea. We can convert everything then.

controls/cGameControlsContext.h
```diff
@@ -17,13 +10,48 @@
...
+#include "controls/mousestates/cMouseRepairState.h"
+
+enum eMouseState {
+    MOUSESTATE_SELECT, // mouse is in 'normal' mode, it is able to select units
+    MOUSESTATE_UNITS_SELECTED, // mouse has selected units
+    MOUSESTATE_REPAIR, // mouse is in repair state
+    MOUSESTATE_PLACE, // mouse is going to place a structure
+};
+
+static const char* mouseStateString(const eMouseState &state) {
```
pbomta, line 34:
```suggestion
static std::string mouseStateString(eMouseState state) {
```
Returning a string is better practice than a raw character pointer. Then the caller doesn't need to worry about the pointer being nullptr or pointing to something that's cleaned up or overwritten. And in modern compilers, we can rely on RVO (return value optimization) so that the string is created in-place and there is no copy (or move).

The eMouseState enum is so small, the advice is to pass it by copy. It's smaller than a pointer, so it can be passed more efficiently (in a CPU register, for example). An extensive coverage of that topic is here: https://www.modernescpp.com/index.php/c-core-guidelines-how-to-pass-function-parameters

controls/mousestates/cMouseNormalState.cpp
```diff
@@ -0,0 +1,195 @@
+#include <algorithm>
+#include "d2tmh.h"
```
pbomta, line 2: It's better to include only what is used in this file. Start with `#include "mousestates/cMouseNormalState.h"` and then add what's missing.

pbomta (approved):
I've mainly commented on C++ style things.
I don't know the codebase well enough to check the code itself. I would say that the architecture improves, though. Putting an abstraction layer around button presses and mouse clicks looks good.

observers/cInputObserver.h
stefanhendriks: I would prefer a composite over multiple inheritance. But yeah I was lazy and make one depend on the other, while this is not logical to do.

include/enums.h
stefanhendriks: Oh that is awesome. I did not know there was such a thing! It was really bugging me to have these enums clash with each other. `enum classes` will solve that for sure. Thanks!

controls/mousestates/cMouseState.h
stefanhendriks: 👍

include/sKeyboardEvent.h
stefanhendriks: thx!

cGame_logic.cpp
stefanhendriks: In this case I simply moved a function, so I did not really bother rewriting it. But, by all means, I do want to learn/improve. So if you have a (different/better) implementation for this (via a small PR?), I am very interested.

controls/cGameControlsContext.h
stefanhendriks: Ah this is good stuff. I kind of try to make it my best practice to have `const <type> & var` in method calls. So I guess this depends on the size of `type` then. 

About `std::string` thx! :)

controls/mousestates/cMouseNormalState.cpp
stefanhendriks: this was laziness from my part. I'll try to get it working without `d2tmh.h`

stefanhendriks: Well, I try to get through this mess, since I need to include `cPlayer.h` somewhere, it brings in the whole world and pain. I'll try to spend a bit more time on it, but if it is too hard, I will delegate that to a separate PR/ticket.

stefanhendriks:
@pbomta thanks for reviewing. I try my best to improve the codebase over time. But I also try to keep things moving (ie, working on new features). It is a balancing act which I'm getting better at over time. I appreciate your comments, I learned a bunch of things from them.

I do wonder what you think of my solution to get it to compile without `d2tmh.h` in the `cNormalMouseState.cpp` file. It felt I had to hack stuff in the `structh.h` file, something weird with `stdint.h` or something.. Ideally I would just include `allegro/gfx.h` and be done with it, but I ran into issues.

controls/mousestates/cMouseNormalState.cpp
pbomta: Sure. This PR is already pretty big. If it results in changing multiple files, we can just add this tech debt.

controls/mousestates/cMouseNormalState.cpp
```diff
@@ -1,8 +1,25 @@
...
+#include "gameobjects/units/cUnit.h"
+#include "gameobjects/structures/cAbstractStructure.h"
+#include "player/cPlayer.h"
+#include "sidebar/cSideBar.h"
+
+#include "cGame.h"
+
+// instead of using d2tmc.h
+extern cGame		  game;
+extern cUnit          unit[MAX_UNITS];
```
pbomta, line 16: 🤔 Although it saves including d2tmc.h, I wonder if it's worth it. If we ever clean up that file (by moving data into the game class, for instance), then we need to touch these lines too and we may forget. I hope we get linker errors by then, but it makes me nervous to declare variables in multiple places. :)

controls/mousestates/cMouseNormalState.cpp
```diff
@@ -1,8 +1,25 @@
...
 
 #include "cMouseNormalState.h"
```
pbomta, line 18: A tip when working with includes: include the header that belongs to the cpp file (so this one) first. Then when compiling cMouseNormalState.cpp, it's checked if the cMouseNormalState.h header can be included separately, so without including other stuff first. If we do this well, then inclusion order no longer matters.

PR #382, opened 2021-12-29
Enabled -Wall and fixes the last remaining warnings

pbomta:


gameobjects/units/cUnit.cpp
```diff
@@ -246,7 +246,7 @@ void cUnit::createExplosionParticle() {
         }
     }
 
-    if (iType == SIEGETANK || iType == DEVASTATOR && rnd(100) < 25) {
+    if ((iType == SIEGETANK || iType == DEVASTATOR) && rnd(100) < 25) {
```
pbomta, line 249: @stefanhendriks, I wonder if I put the parentheiss correctly. It seems this is what was intented, but it is different from what was executed. The randomness was only applied to the devastator. Do you know what this line should do?

stefanhendriks: I think you found a bug here and correctly fixed it. In fact, I doubt the `SIEGEDIE` particle should appear for a `DEVASTATOR`, but alas, that was implemented so lets keep that for sure for now.

include/enums.h
```diff
@@ -173,7 +173,7 @@ namespace buildOrder {
 		REMOVEME
 	};
 
-    static const char* eBuildOrderStateString(const eBuildOrderState &state) {
+    inline const char* eBuildOrderStateString(const eBuildOrderState &state) {
```
stefanhendriks, line 176: what is the difference between `static` and `inline` here?

player/brains/missions/cPlayerBrainMissionKind.h
```diff
@@ -60,25 +60,8 @@ namespace brains {
...
-            case ePlayerBrainMissionKind::PLAYERBRAINMISSION_KIND_FIND_SPICE: return "PLAYERBRAINMISSION_KIND_FIND_SPICE";
-            case ePlayerBrainMissionKind::PLAYERBRAINMISSION_KIND_HARASS: return "PLAYERBRAINMISSION_KIND_HARASS";
-            default:
-                assert(false && "Unknown kind?");
-                break;
-        }
-        return "";
-    }
-
+    const char* ePlayerBrainMissionKindString(const ePlayerBrainMissionKind &kind);
```
stefanhendriks, line 63: here it is not `inline` ?

stefanhendriks:
a few questions, but other than that lgtm

player/brains/missions/cPlayerBrainMissionKind.h
pbomta: Because this is just a declaration. The implementation (definition) of the function is in a .cpp file. In this case, inline makes no sense, as it only applies to functions with the function body in the header file.

include/enums.h
pbomta: For free functions (so not member functions), static means that it has "internal linkage", which means it is only accessible from the current compilation unit. So it only makes sense for definitions in the .cpp file. As a side-note: putting symbols in an anonymous namespace has the same effect and is recommended above the static keyword.

The inline keyword means that multiple implementations of a function may exist and that the linker can just pick one. This is for functions with the implementation in the header file.

PR #383, opened 2021-12-29
build: update includes for cMouseNormalState

pbomta:
This is what I would do. What do you think?

### DIFF
--- a/controls/mousestates/cMouseNormalState.cpp
+++ b/controls/mousestates/cMouseNormalState.cpp
@@ -1,7 +1,6 @@
-#include <algorithm>
+#include "cMouseNormalState.h"
 
 #include "controls/cGameControlsContext.h"
-#include "controls/cMouse.h"
 
 #include "utils/cRectangle.h"
 #include "gameobjects/units/cUnit.h"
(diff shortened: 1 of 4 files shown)


controls/mousestates/cMouseNormalState.cpp
```diff
@@ -1,25 +1,21 @@
...
-// instead of using d2tmc.h
-extern cGame		  game;
-extern cUnit          unit[MAX_UNITS];
-
-#include "cMouseNormalState.h"
-
 #include "data/gfxaudio.h"
 #include "data/gfxdata.h"
 
+#include <algorithm>
```
stefanhendriks, line 17: so, isn't this a std include? Wouldn't it make sense to have this one at top? I always assumed it was:
- standard lib includes
- extern lib includes
- my includes

pbomta: Ah, yes, I guess habits from my work-life creep in here. :)
The only important thing about include ordering is that the header file that belongs to the .cpp file (cMouseNormalState.h) is included first. Then, when compiling this file, the header will be checked that it can be included and used as-is, so without including other files first.

What comes after that, anything goes. I believe the Google Style Guide for C++ (https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) recommends: Related header, C system headers, C++ standard library headers, other libraries' headers, your project's headers.
At my job, we use: Related header, your project's headers, other libraries' headers, standard library headers.

PR #384, opened 2022-01-01
refactor: eliminate a double query of mouse coordinates

pbomta:
Found this when fixing "unused-parameter" warnings. Playtested it and this works.


stefanhendriks:
good catch; and it possibly is duplicated by new `gui` objects that were created lately, so that can be simplified further later.

PR #389, opened 2022-01-01
chore: enable missing virtual destructor warning

pbomta:


stefanhendriks:
lgtm.

the `abstractUnit` was brought up and never used sadly. But who knows it might in the future.

stefanhendriks:
(it would be even fine to remove it)

PR #390, opened 2022-01-01
chore: eliminates useless casts

pbomta:
A nice warning to add. It's not part of Wextra for some reason, but IMHO it helps to get clean code.

### DIFF
--- a/player/cPlayerDifficultySettings.cpp
+++ b/player/cPlayerDifficultySettings.cpp
@@ -1,6 +1,7 @@
 #include "../include/d2tmh.h"
 #include "cPlayerDifficultySettings.h"
 
+#include <cmath>
 
 cPlayerDifficultySettings::cPlayerDifficultySettings() :
                                 m_moveSpeedFactor(1.0f),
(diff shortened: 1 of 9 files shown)


player/cPlayerDifficultySettings.cpp
```diff
@@ -27,23 +28,23 @@ cPlayerDifficultySettings::~cPlayerDifficultySettings() {
 }
 
 float cPlayerDifficultySettings::getMoveSpeed(int iUnitType, int slowDown) {
-    return (int)(sUnitInfo[iUnitType].speed) + ((float)slowDown * m_moveSpeedFactor);
+    return sUnitInfo[iUnitType].speed + slowDown * m_moveSpeedFactor;
 }
 
 float cPlayerDifficultySettings::getBuildSpeed(int iSpeed) {
-    return (int) (iSpeed * m_buildSpeedFactor);
+    return std::floor(iSpeed * m_buildSpeedFactor);
```
pbomta, line 35: Was this the intent? If yes, this makes it more explicit. If no, we found a bug. :)

PR #391, opened 2022-01-01
refactor: enable unused parameter warning

pbomta:
The warning is part of -Wextra, which we should also enable in time.
I also did some collateral cleanup (constness and pragma once).

### DIFF
--- a/gameobjects/units/cUnitUtils.cpp
+++ b/gameobjects/units/cUnitUtils.cpp
@@ -1,6 +1,6 @@
 #include "../../include/d2tmh.h"
 
-cUnitUtils *cUnitUtils::instance = NULL;
+cUnitUtils *cUnitUtils::instance = nullptr;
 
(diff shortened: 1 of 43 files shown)


gameobjects/units/cUnitUtils.cpp
```diff
@@ -22,13 +22,13 @@ int cUnitUtils::findUnit(int type, int iPlayerId) {
...
-				if (unit[i].iPlayer == iPlayerId) {
-					return i;
-				}
-			}
-		}
+    if (i == iIgnoreUnitId || !unit[i].isValid()) {
+      continue;
+    }
+
+    if (unit[i].iPlayer == iPlayerId && unit[i].iType == type) {
```
pbomta, line 29: Here, I decided to use the type parameter. 
This function isn't used (neither is the class), but I didn't want to throw away work in case this is the direction we want to take with the split-up of the large cUnit.cpp.

PR #392, opened 2022-01-03
introduce libfmt version 8.0.1 and convert ini.cpp and cUnit.cpp to use it

pbomta:
Tested manually in debug mode. The logging seems fine.
I didn't convert everything yet, that would make the PR too big.


stefanhendriks:
works great, thanks! awesome to have it in, thanks!

PR #394, opened 2022-01-05
fix designated and structured initializers to conform to C++20

pbomta:
A mechanical change.
`(type) { a: something, b: something)` turns out to be nonstandard.
The standard form is: `type{ a = something, b = something }`
This is needed to enable C++20 officially, which we are already using with these initializers.


stefanhendriks (approved):
Checked, looks like a lot of work (or l33t regex skills) to make this happen.

Also compiles fine on my machine.

So lgtm! thanks

pbomta:
> Checked, looks like a lot of work (or l33t regex skills) to make this happen.

It wasn't too bad. Find & replace in VSCode did the trick. :) Luckily, the search pattern was pretty unique.

PR #395, opened 2022-01-05
sets the C++ standard to 20

pbomta:
Using ++ (or +=) operators on volatile ints is no longer allowed with C++20, because it doesn't work properly. There is more information to be found on this, but the replacement is pretty simple: a = a + 1.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.21)
 
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
 
 project(d2tm)
 
--- a/cGame_logic.cpp
+++ b/cGame_logic.cpp
@@ -841,8 +841,8 @@ bool cGame::setupGame() {
 
 	// Install timers
 	install_int(allegro_timerunits, 100); // 100 milliseconds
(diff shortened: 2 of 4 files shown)


CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.21)
 
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
```
stefanhendriks, line 3: I don't know if MinGW is up to date with this, atleast not for std (I have the latest, and the latest std is v17). So not sure if this is just MinGW being outdated (although it seems to compile, when I look at the GH action).

Is there an std v20?

stefanhendriks:
lgtm, just one question before merging 👍

CMakeLists.txt
pbomta: Interesting... I have:
`g++.exe (Rev5, Built by MSYS2 project) 11.2.0`
it supports `-std=c++20`. According to the GCC documentation, that flag exists in GCC version 10 and higher.
Do you use MSYS2 as well?

stefanhendriks: (I use `git bash` btw, I run this on a windows machine). Not sure about MSYS2 btw. I just install MinGW. Not sure if that uses MSYS in anyway.

stefanhendriks:
I had to install msys2 and all that jazz to get this to compile, but it works flawlessly.

PR #396, opened 2022-01-05
remove redundant tags

pbomta:
I'm having fun with these extra warnings. :)
Saves us from running cppcheck or other static checkers.

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -20,7 +20,7 @@ jobs:
         with:
           msystem: MINGW32 # D2TM is built within a MinGW32 bit environment
           update: true
-          install: mingw-w64-i686-gcc mingw-w64-i686-binutils mingw-w64-i686-cmake mingw-w64-i686-ninja git
+          install: mingw-w64-i686-gcc mingw-w64-i686-binutils mingw-w64-i686-cmake mingw-w64-i686-make git
       - name: Checkout
         uses: actions/checkout@v2
         with:
@@ -30,8 +30,8 @@ jobs:
           cd "${{ github.workspace }}"
           mkdir build
           cd build
-          cmake .. -G "Ninja"
+          cmake .. -G "MinGW Makefiles"
       - name: Build
         run: |
           cd "${{ github.workspace }}\build"
-          ninja
+          make -j -l 3
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -20,7 +20,7 @@ jobs:
         with:
           msystem: MINGW32 # D2TM is built within a MinGW32 bit environment
           update: true
-          install: mingw-w64-i686-gcc mingw-w64-i686-binutils mingw-w64-i686-cmake mingw-w64-i686-make git
+          install: mingw-w64-i686-gcc mingw-w64-i686-binutils cmake make git
       - name: Checkout
         uses: actions/checkout@v2
         with:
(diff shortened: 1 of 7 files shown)


utils/cSeedMapGenerator.h
```diff
@@ -37,38 +37,38 @@ class cSeedMapGenerator {
 
 	/* these is array used to find corresponding points */
 	/*
-	      A�B�C    00��20��40
-	      �����    ����������
-	      D�E�F    02��22��42
-	      �����    ����������
-	      G�H�I    04��24��44
+	      A�B�C    00��20��40
```
stefanhendriks, line 40: what happened here? :)

stefanhendriks:
Maybe the GH action does not support this? I see:
```g++.exe: error: unrecognized command line option '-Wredundant-tags'; did you mean '-Wredundant-decls'?```

File: utils/cSeedMapGenerator.h (hunk 1 above)
Comment (pbomta) [reply utils/cSeedMapGenerator.h:40]: Oh, I didn't notice that! Let's not merge this as-is then.
I did see some messages on line breaks and carriage returns being converted from windows style to unix style, so perhaps that's where it went wrong...

**Comment by pbomta** [comment]
> Maybe the GH action does not support this? I see: `g++.exe: error: unrecognized command line option '-Wredundant-tags'; did you mean '-Wredundant-decls'?`

There seems to be a difference between my local environment and what is used on GH. I'll look into it.

File: utils/cSeedMapGenerator.h (hunk 1 above)
Comment (pbomta) [reply utils/cSeedMapGenerator.h:40]: The characters were unicode dots (<B7>), not ASCII dots. My VSCode messed them up, I think. I've replaced them with ASCII dots.

**Comment by stefanhendriks** [comment]
I do notice the GH action uses mingw v8.1.0 , (there seems to be a 9.x version).

File: .github/workflows/build_cmake.yml [hunk 2]
```diff
@@ -8,38 +8,30 @@ on:
...
-      - name: Running cmake (windows)
+      - name: Configure
         run: |
-          cd ${{ github.workspace }}
+          cd "${{ github.workspace }}"
           mkdir build
           cd build
-          cmake .. -G "MinGW Makefiles"
-
+          cmake .. -G "Ninja"
```
stefanhendriks, line 33: Is there a reason we compile under ninja here? Or perhaps I don't understand MinGW/Ninja stuff properly (most probably). I thought using the `MinGW Makefiles` one was needed when using MinGW. But perhaps not?

stefanhendriks:
Nice to see the GH check is 2x faster! 🎉

.github/workflows/build_cmake.yml
pbomta: My colleagues told me ninja builds faster. I took that knowledge at face value. :) Both tools can be used with MinGW.
But now I googled a bit and two sources that I found say that there is no performance difference. Perhaps I should put make back.

.github/workflows/build_cmake.yml
```diff
@@ -8,38 +8,30 @@ on:
...
           cd build
           cmake .. -G "MinGW Makefiles"
-
       - name: Build
         run: |
-          cd build
-          cmake --build . --target all -- -j 6
\ No newline at end of file
+          cd "${{ github.workspace }}\build"
+          cmake --build . --target all -- -j -l 3
```
stefanhendriks, line 37: I can't find that `-l` does, but it looks like you want to have it run at 3 threads (instead of 6) ?

pbomta: I had to look that up before I wrote it. I didn't want to just specify the number of parallel builds, as we don't know on what kind of system it runs. This could also change in the future. What this does instead is:
-j -> run in parallel, unlimited
-l 3 -> spawn new compilation processes if the average load of the system is lower than 3.
Looked like a nice thing to do to me.

Come to think of it, I don't know what Ninja does. I believe it queries the number of cores of the system.

PR #397, opened 2022-01-08
convert cPlayer to use libfmt for logging

pbomta:
Also found a tweak for Allegro: defining HAVE_STDINT_H (which we indeed have) solves compilation errors when certain headers (such as format/format.h) are included after Allegro has been included. I think Allegro relies on a config script that generates a header with these defines in it. I've seen several such (mostly older) projects use that technique.


stefanhendriks:
nice find about `HAVE_STDINT_H `. I did notice it being mentioned somewhere, but didn't know if I could just define it and be done with it.

cGame.h
```diff
@@ -11,11 +11,12 @@
...
 #pragma once
 
-#include <controls/cMouse.h>
-#include <data/cAllegroDataRepository.h>
-#include <observers/cScenarioObserver.h>
-#include <utils/cRectangle.h>
-#include <utils/cSoundPlayer.h>
+#include "controls/cMouse.h"
+#include "data/cAllegroDataRepository.h"
+#include "definitions.h"
```
stefanhendriks, line 16: I feel like this file also might be candidate for removal some time? (ie, we could move things from within `definitions.h` closer to their classes where they are used?). For now 👌 though.

player/cPlayer.cpp
```diff
@@ -1,8 +1,13 @@
...
+
+#include "cPlayer.h"
+
+#include "utils/common.h"
+
+#include <fmt/format.h>
+
 #include <vector>
 #include <algorithm>
-#include "../include/d2tmh.h"
```
stefanhendriks, line 3: 🎉

player/cPlayer.h
```diff
@@ -51,7 +53,7 @@ struct s_PlaceResult {
 class cPlayer : public cScenarioObserver {
 
 public:
-    static const int MAX_STRUCTURE_BMPS = MAX_STRUCTURETYPES*2;
+    static constexpr int MAX_STRUCTURE_BMPS = MAX_STRUCTURETYPES*2;
```
stefanhendriks, line 56: what is the difference between `const` and `constexpr` ? A quick google tells me it has to do with when this is initialized (compile time vs runtime). Thats it? Is it a best practice to use this instead of `const` ?

player/cPlayerNotification.cpp
```diff
@@ -11,11 +12,11 @@ const char* eNotificationTypeString(const eNotificationType &type) {
             assert(false && "Unknown eNotificationType?");
             break;
     }
-    return "";
+    return {};
 }
 
 cPlayerNotification::cPlayerNotification(std::string msg, eNotificationType type) {
-    this->msg = msg;
+    this->msg = std::move(msg);
```
stefanhendriks, line 19: aha, this does seem to be something you have to be aware of if you instantiate a `cPlayerNotification`? Or... wait, it probably copies `msg` already from the constructor args, then you simply move from there, thats what you intent correct?

stefanhendriks (approved):

cGame.h
pbomta: Certainly. That's a direction of improvement that's also on my mind.

player/cPlayer.h
pbomta: constexpr Allows variables and functions to be evaluated at compile-time. In this case, the `MAX_STRUCTURETYPES*2` is calculated by the compiler and not by the program when it's initialized. Doesn't make that much difference in this case, but it is indeed a best practice. Because we can now use this variable in other constexpr functions/expressions and they will then also be evaluated at compile-time.

pbomta: I also think the optimizer can throw the variable out if it's not used at runtime anymore. So there would be no memory reserved for this integer. Not 100% sure, though. (As one can never be with the optimizer. ;) )

player/cPlayerNotification.cpp
pbomta: I started typing an explanation, but it's quite involved. Perhaps a better fix is to change the constructor of `cPlayerNotification` to take a string by const-ref. Because what was there copied the string twice.

PR #398, opened 2022-01-08
remove the fblend library and its usage

### LINKED ISSUE #387: Remove FBLEND dependency

pbomta:
Resolves #387 

Also moves fading from the time manager to the game logic. The game logic may not be the best final place for it, but it certainly doesn't belong to the time manager. :)

Used a neat trick where the custom deleter of a unique_ptr is used to destroy temporary bitmaps. With this technique, the temporary bitmap is always destroyed when going out of scope and one can never forget to destroy it.

Playtesting showed no issues. Also no performance issues, but that was not expected anyway on modern hardware.


CMakeLists.txt
```diff
@@ -76,7 +76,6 @@ target_include_directories(${PROJECT_NAME}
 
 target_link_libraries(${PROJECT_NAME}
   PRIVATE
-    fblend # Must link before alleg
```
stefanhendriks, line 79: the first step to get rid of Allegro, yay!

cGame_logic.cpp
```diff
@@ -17,6 +17,13 @@
 
 #include <fmt/core.h>
 
+namespace {
+
+constexpr auto kMinAlpha = 0;
```
stefanhendriks, line 22: why the prefix `k`?  (just curious)

cGame_logic.cpp
```diff
@@ -547,38 +554,41 @@ void cGame::shakeScreenAndBlitBuffer() {
...
-			offset = 9;
+    TIMER_shake = std::min(TIMER_shake, 69);
+    int offset = std::min(TIMER_shake / 5, 9);
 
 		shake_x = -abs(offset/2) + rnd(offset);
 		shake_y = -abs(offset/2) + rnd(offset);
 
-		blit(bmp_screen, bmp_throttle, 0, 0, 0+shake_x, 0+shake_y, screen_x, screen_y);
+		blit(bmp_screen, bmp_throttle, 0, 0, 0 + shake_x, 0 + shake_y, screen_x, screen_y);
+    vsync();
```
stefanhendriks, line 572: if you remove vsync we get more fps. No need to worry about tearing due to double buffering technique.

cGame_logic.cpp
```diff
@@ -547,38 +554,41 @@ void cGame::shakeScreenAndBlitBuffer() {
...
-        } else {
-			BITMAP *temp = create_bitmap(game.screen_x, game.screen_y);
-			assert(temp != NULL);
-			clear(temp);
-			fblend_trans(bmp_screen, temp, 0, 0, fadeAlpha);
-			blit(temp, screen, 0, 0, 0, 0, screen_x, screen_y);
-			destroy_bitmap(temp);
+		if (fadeAction == eFadeAction::FADE_NONE) {
+  		// Not shaking and not fading.
+      vsync();
```
stefanhendriks, line 579: same

cGame_logic.cpp
```diff
@@ -547,38 +554,41 @@ void cGame::shakeScreenAndBlitBuffer() {
...
+    } else {
+      // Fading
+      assert(fadeAlpha >= kMinAlpha);
+      assert(fadeAlpha <= kMaxAlpha);
+      auto temp = std::unique_ptr<BITMAP, decltype(&destroy_bitmap)>(create_bitmap(game.screen_x, game.screen_y), destroy_bitmap);
+			assert(temp);
+			clear(temp.get());
+      set_trans_blender(0, 0, 0, fadeAlpha);
+      draw_trans_sprite(temp.get(), bmp_screen, 0, 0);
+      vsync();
```
stefanhendriks, line 590: same

cGame_logic.cpp
```diff
@@ -547,38 +554,41 @@ void cGame::shakeScreenAndBlitBuffer() {
...
-			destroy_bitmap(temp);
+		if (fadeAction == eFadeAction::FADE_NONE) {
+  		// Not shaking and not fading.
+      vsync();
+      blit(bmp_screen, screen, 0, 0, 0, 0, screen_x, screen_y);
+    } else {
+      // Fading
+      assert(fadeAlpha >= kMinAlpha);
+      assert(fadeAlpha <= kMaxAlpha);
+      auto temp = std::unique_ptr<BITMAP, decltype(&destroy_bitmap)>(create_bitmap(game.screen_x, game.screen_y), destroy_bitmap);
```
stefanhendriks, line 585: so, my c++ is not as proficient as yours.. let me try to unpack this:
the `std::unique_ptr` will trigger `destroy_bitmap` once the `temp` variable goes out of scope. And you instantiate it via the `create_bitmap` within its parentheses?
I do see `destroy_bitmap` twice in that line. 

Hmm. So I read at https://en.cppreference.com/w/cpp/memory/unique_ptr, but that is still a bit too much info for me. I also see `decltype` used, never saw that (did read https://en.cppreference.com/w/cpp/language/decltype though). I trust you on this 👍

drawers/cAllegroDrawer.h
```diff
@@ -29,29 +37,30 @@ class cAllegroDrawer {
...
-        void drawRectangle(BITMAP *dest, int x, int y, int width, int height, int color);
-        void drawRectangle(BITMAP *dest, const cRectangle &pRectangle, int color);
-        void drawRectangleFilled(BITMAP *dest, const cRectangle &pRectangle, int color);
+    void drawRectangle(BITMAP *dest, int x, int y, int width, int height, int color);
+    void drawRectangle(BITMAP *dest, const cRectangle &pRectangle, int color);
+    void drawRectangleFilled(BITMAP *dest, const cRectangle &pRectangle, int color);
+    void drawRectangleTransparentFilled(BITMAP *dest, const cRectangle& rect, int color, int alpha);
 
-        int getColor_BLACK() { return colorBlack; }
+    int getColor_BLACK() { return colorBlack; }
```
stefanhendriks, line 53: I do notice we have different tab/spaces/indenting settings. It might be worth setting up some kind of editorconfig and let our IDE's sync with that to make the diffs less noisy.

drawers/cAllegroDrawer.cpp
```diff
@@ -200,6 +202,18 @@ void cAllegroDrawer::drawRectangleFilled(BITMAP *dest, const cRectangle &pRectan
     rectfill(dest, pRectangle.getX(), pRectangle.getY(), pRectangle.getEndX(), pRectangle.getEndY(), color);
 }
 
+void cAllegroDrawer::drawRectangleTransparentFilled(BITMAP *dest, const cRectangle& rect, int color, int alpha) {
+  assert(alpha >= 0);
+  assert(alpha <= 255);
+
+  // Create a temporary bitmap that automatically destroys itself when going out-of-scope
+  std::unique_ptr<BITMAP, decltype(&destroy_bitmap)> rect_bmp{create_bitmap(rect.getWidth(), rect.getHeight()), destroy_bitmap};
```
stefanhendriks, line 210: nice intermediate solution.

I was thinking if we could optimize this. Ie, I notice the rectangle (bitmap) we create is often the same dimension(s). So we might even be able to cache these created bitmaps for re-usage (to re-draw) and only destroy them once the `cAllegroDrawer` gets destroyed.

It saves create/deletion every frame. 

So basically we get a map I suppose, where the key could be `<width>x<height>` and the value the `BITMAP *`?

drawers/cOrderDrawer.cpp
```diff
@@ -17,20 +18,19 @@ cOrderDrawer::~cOrderDrawer() {
 }
 
 void cOrderDrawer::drawOrderPlaced() {
-	BITMAP *bmp_trans=create_bitmap(buttonBitmap->w, buttonBitmap->h);
-	clear_to_color(bmp_trans, makecol(255,0,255));
+  // Create a temporary bitmap that destroys itself when going out of scope
+  std::unique_ptr<BITMAP, decltype(&destroy_bitmap)> bmp_trans{create_bitmap(buttonBitmap->w, buttonBitmap->h), destroy_bitmap};
```
stefanhendriks, line 22: lets move this line into the constructor so we can re-use the bitmap, and only destroy it once we destroy `cOrderDrawer` ?

utils/cTimeManager.cpp
```diff
@@ -115,20 +115,7 @@ void cTimeManager::handleTimerAllegroTimerSeconds() {
...
-            }
-        } else if (game.fadeAction == eFadeAction::FADE_IN) {
-            game.fadeAlpha += 2;
-            if (game.fadeAlpha > 255) {
-                game.fadeAlpha = 255;
-                game.fadeAction = eFadeAction::FADE_NONE;
-            }
-        }
-
+        game.think_fading();
```
stefanhendriks, line 118: nice to see the same naming strategy, I would have done the same 👍

stefanhendriks: oh, 🤦‍♂️ - I see this function already existed and just got moved... :)

cGame_logic.cpp
```diff
@@ -1335,8 +1345,23 @@ void cGame::setState(int newState) {
 }
 
 void cGame::think_fading() {
+    // Fading of the entire screen
+    if (fadeAction == eFadeAction::FADE_OUT) {
```
stefanhendriks, line 1349: it is fine to have it here for now 👍

stefanhendriks (approved):
A few nitpicks, would be nice to have some caching going on.

All in all very pleased to see fblend go. I did not playtest this (yet), as my `data` files are out of sync and it crashes. (and I don't want to lose my data files yet, so I am lazy, sorry :) ).

cGame_logic.cpp
pbomta: Oh, that's my work coding guidelines creeping in. We use that to signify it's a constant that is the same for the whole program duration. I believe it comes from the Google C++ Style Guide. See: https://google.github.io/styleguide/cppguide.html#Constant_Names

Now, I don't think the Google Style Guide is the be all and end all of style guides, but this one seems useful to me.

drawers/cAllegroDrawer.cpp
pbomta: Not a bad idea, actually.

PR #401, opened 2022-01-11
remove cSimpleCalculator in favor of built-in maths

pbomta:
Came across this when working on something else. A small and easy cleanup.


stefanhendriks:
yeah looks good. Don't know why I thought it was a good idea to have this `cSimpleCalculator`. Thanks! :)

PR #403, opened 2022-01-13
Refactor logger to be a self-sufficient singleton

pbomta:
Now it can be used from other singletons too. This is useful if we turn cSoundPlayer into a singleton, for example.
The logger initializes itself now and clears the logfile.

### DIFF
--- a/utils/cLog.cpp
+++ b/utils/cLog.cpp
@@ -44,10 +44,10 @@ std::string getLogComponentString(eLogComponent component) {
             return std::string("BULLET");
         case COMP_AI:
             return std::string("AI");
-    case COMP_UPGRADE_LIST:
-      return std::string("UPGRADE_LIST");
-      case COMP_BUILDING_LIST_UPDATER:
-      return std::string("BUILDING_LIST_UPDATER");
+        case COMP_UPGRADE_LIST:
+            return std::string("UPGRADE_LIST");
+        case COMP_BUILDING_LIST_UPDATER:
+            return std::string("BUILDING_LIST_UPDATER");
         case COMP_MAP:
             return std::string("MAP");
         case COMP_SIDEBAR:
@@ -83,12 +83,13 @@ std::string getLogOutcomeString(eLogOutcome outcome) {
             return std::string("FAILED");
         case OUTC_NONE:
             return std::string("NONE");
-      case OUTC_UNKNOWN:
-      break;
-      case OUTC_IGNOREME:
-      break;
+        case OUTC_UNKNOWN:
+            break;
+        case OUTC_IGNOREME:
+            break;
     }
-     return std::string("UNIDENTIFIED");
+
+    return std::string("UNIDENTIFIED");
 }
 
 std::string getLogHouseString(int houseId) {
@@ -114,9 +115,9 @@ std::string getLogHouseString(int houseId) {
 }
 
 cLogger* cLogger::getInstance() {
-  // The logger is initialized once as soon as this function is called for the first time.
-  static cLogger theLogger;
-  return &theLogger;
+    // The logger is initialized once as soon as this function is called for the first time.
+    static cLogger theLogger;
+    return &theLogger;
 }
 
 void cLogger::log(eLogLevel level, eLogComponent component, const std::string& event, const std::string& message) {
@@ -149,8 +150,8 @@ void cLogger::log(eLogLevel level, eLogComponent component, const std::string& e
     logline += "|";
 
     if (playerId >= GENERALHOUSE) {
-            logline += getLogHouseString(houseId);
-            logline += "|";
+        logline += getLogHouseString(houseId);
+        logline += "|";
     }
 
     if (playerId >= HUMAN) {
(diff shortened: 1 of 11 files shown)


utils/cLog.cpp
```diff
@@ -1,262 +1,204 @@
-#include "../include/d2tmh.h"
 #include "cLog.h"
 
+#include "definitions.h"
+
+#include <cstdio>
+#include <ctime>
+#include <system_error>
+
+extern bool	bDoDebug;
```
pbomta, line 9: I put this one here, instead of including the entire header file with all the globals.
It's not ideal, but for now, I think this is a step forward.

utils/cLog.cpp
```diff
@@ -1,262 +1,204 @@
...
-	strcpy(str, line.c_str());
-
-	logCommentLine(str);
-	logCommentLine(txt);
-	logCommentLine(str);
-
-	delete[] str;
+cLogger* cLogger::getInstance() {
+    // The logger is initialized once as soon as this function is called for the first time.
+    static cLogger theLogger;
```
pbomta, line 119: Here it is! No longer a pointer, but just a static variable. C++ Guarantees that this variable will be initialized if the getInstance function is called for the first time. Destruction order of globals in C++ is the inverse of construction order.
If other singletons need it, they can call this getInstance in the constructor and then it is guaranteed it will be initialized before the other singleton and destroyed after the other singleton.

utils/cLog.cpp
```diff
@@ -1,262 +1,204 @@
...
-			return std::string("MERCENARY");
-		default:
-			return std::string("UNKNOWN HOUSE");
-	}
-	return std::string("UNIDENTIFIED");
-}
-
-
-// courtesy from : http://www.codeguru.com/forum/showthread.php?t=477894
-std::string cLogger::getCurrentFormattedTime() {
```
pbomta, line 126: If we want this functionality in the future, I can recommend the standard chrono library.

stefanhendriks:
Thx for the PR, checked it with mobile. Will check later with PC. But looks good so far!

utils/cLog.h
```diff
@@ -1,48 +1,66 @@
-#ifndef D2TM_CLOG_H_
-#define D2TM_CLOG_H_
+#pragma once
 
-class cLogger {
-
-public:
-	void log(eLogLevel level, eLogComponent component, const std::string& event, const std::string& message);
-	void log(eLogLevel level, eLogComponent component, const std::string& event, const std::string& message, eLogOutcome outcome, int playerId, int houseId);
```
stefanhendriks, line 8: this is interesting, so you moved this out of this class. And you put these in the anonymous namespace. I would've expected it to be moved to a `private:` section in the `cLogger` class. What is the rationale not to do this?

utils/cLog.cpp
```diff
@@ -1,262 +1,204 @@
...
 
+#include "definitions.h"
+
+#include <cstdio>
+#include <ctime>
+#include <system_error>
+
+extern bool	bDoDebug;
+
+namespace {
```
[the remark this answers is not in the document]
stefanhendriks: How /when do you decide to do this as opposed to have methods within the `private:` part of a class?

utils/cLog.cpp
```diff
@@ -1,262 +1,204 @@
...
+
+std::string getLogLevelString(eLogLevel level) {
+    // LOG_TRACE, LOG_WARN, LOG_ERROR, LOG_FATAL
+    switch (level) {
+        case LOG_FATAL:
+            return std::string("FATAL");
```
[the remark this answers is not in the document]
stefanhendriks: I picked this up for you. :)

utils/cLog.cpp
stefanhendriks: Aha, so I see the difference. I do wonder though, is this preferred over `new` ? The difference afaik this is now on the stack (and not the heap). It is initialized once (and only once), in both cases (right?). So is this more like a 'good practice' in C++ land?

stefanhendriks (changes requested):
Please take a look at the log result to make it look like previous version , and preferably have the D2TM header first (before loading the `game.ini`). 

Other than that, a few questions and then good to go :)

utils/cLog.cpp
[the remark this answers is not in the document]
pbomta: If a function does not use any class member variables, it doesn't have to be tied to the object instantiated from the class. Two choices here: static member function, or a non-member function in the translation unit (.cpp file). That depends on if the function is called from outside of the translation unit. If it is: static member function. If not: in an anonymous namespace in the .cpp file. 
The reason is: information hiding. If we hide a function much as we can, it is restricted from where it can be called, which helps local reasoning and understanding the code.

utils/cLog.cpp
pbomta: Yes, it's preferred because we want the language to manage memory for us. I believe the guideline is called "no naked new & delete".
Variables with static storage duration, like this one, are - similar to globals - allocated on a reserved piece of memory. I don't know all the details, but it's neither the stack, nor the heap.
In both cases, it's initialized once, yes.
Doing it this way, guarantees thread safety. Not that we call this currently from different threads, but we could. :)
The pattern is called "Meyers Singleton", after Scott Meyers.
Oh and ideally, we would return a reference here to tell callers that it can never be nullptr. But it's too much for this PR to change all the calling code as well.

utils/cLog.cpp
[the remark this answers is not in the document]
pbomta: Thanks. There is indeed to need to explicitly initialize a string.

pbomta:
> Please take a look at the log result to make it look like previous version , and preferably have the D2TM header first (before loading the `game.ini`).

Yes! Good catch. I made a silly copy-paste error.

PR #405, opened 2022-01-13
Bugfix: Init minimapDrawer upon playing new mission

stefanhendriks:
for #285

### DIFF
--- a/drawers/cMiniMapDrawer.cpp
+++ b/drawers/cMiniMapDrawer.cpp
@@ -2,7 +2,7 @@
 #include "cMiniMapDrawer.h"
 
 
-cMiniMapDrawer::cMiniMapDrawer(cMap *theMap, cPlayer * thePlayer, cMapCamera * theMapCamera) :
+cMiniMapDrawer::cMiniMapDrawer(cMap *theMap, cPlayer *thePlayer, cMapCamera *theMapCamera) :
         m_isMouseOver(false),
         map(theMap),
         player(thePlayer),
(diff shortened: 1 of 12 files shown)


utils/cTimeManager.cpp
```diff
@@ -242,7 +243,7 @@ void cTimeManager::handleTimerUnits() {
                 }
             }
 
-            map.think_minimap();
+            drawManager->think();
```
pbomta, line 246: 👍 Looks more consistent this way.

drawers/cMiniMapDrawer.cpp
```diff
@@ -28,12 +30,20 @@ cMiniMapDrawer::cMiniMapDrawer(cMap *theMap, cPlayer * thePlayer, cMapCamera * t
 }
 
 cMiniMapDrawer::~cMiniMapDrawer() {
-	map = NULL;
-	mapCamera = NULL;
-	iStaticFrame = STAT14;
+    map = NULL;
+    mapCamera = NULL;
```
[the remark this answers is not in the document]
stefanhendriks: good one!

utils/cTimeManager.cpp
stefanhendriks: yes, for now this is fine. I think `drawManager` should eventually be the `combat state` object that does not exist yet. And the drawing of the `combat` (ie, actual playing of the mission) will be what the `drawManager` is now.

PR #408, opened 2022-01-15
remove system and global headers

pbomta:
Reduces the size of the "uber" header file.


include/d2tmh.h
```diff
@@ -1,14 +1,12 @@
...
-#define D2TMH_H
+#pragma once
 
-#include "system.h"
 #include "allegroh.h"
 #include "data.h"
 #include "definitions.h"
 #include "structs.h"
-#include "global.h"
+// #include "global.h"
```
stefanhendriks, line 9: can be removed I guess :)

stefanhendriks (approved):
Nice to see this. It looks like it becomes easier and easier? Although some more hard changes will come up ahead. 

A nice improvement indeed :)

include/d2tmh.h
pbomta: Oops! Good catch.

pbomta:
> Nice to see this. It looks like it becomes easier and easier? Although some more hard changes will come up ahead.
> 
> A nice improvement indeed :)

Thanks!
It's indeed slowly getting easier.

PR #409, opened 2022-01-16
add an object to manage Allegro init and de-init

pbomta:
Also fixes a bug in the logger that resulted in an empty log file when the game crashes or is forcefully killed.

So as soon as cGame is no longer a global, but created in the main function, we can constuct this object in cGame's constructor and it will be handled automatically.
The other plan with this is to pass the init object to users of Allegro, such as cSoundPlayer. Then it is programmatically proven that Allegro is initialized before the sound player is constructed.


utils/cPlatformLayerInit.cpp
```diff
@@ -0,0 +1,38 @@
...
+#include <stdexcept>
+
+static bool allegroInitialized = false;
+
+cPlatformLayerInit::cPlatformLayerInit(const std::string& path_to_config_file) {
+    cLogger *logger = cLogger::getInstance();
+    logger->logHeader("Allegro");
+
+    if (allegroInitialized) {
+        throw std::runtime_error("Allegro initialization was attempted twice.");
```
stefanhendriks, line 16: this is something new (I believe there is no other code path throwing any kind of exceptions). What is the consensus about this? Afaik exceptions throwing/catching is a no-no in C++ land?

cGame_logic.cpp
```diff
@@ -653,11 +654,8 @@ void cGame::shutdown() {
 
     logbook("Allegro MP3 library shut down.");
 
-    // Now we are all neatly closed, we exit Allegro and return to OS.
-    allegro_exit();
-
-    logbook("Allegro shut down.");
-    logbook("Thanks for playing.");
+    // Release the game dev framework, so that it can do cleanup
+    _PLInit.reset();
```
stefanhendriks, line 658: I am not 100% sure about how unique ptr works,  but afaik this basically does a `delete` (nothing to replace it with), correct?

I'll look into unique_ptr's more myself in a playground project I have.

cGame_logic.cpp
```diff
@@ -706,16 +704,9 @@ bool cGame::setupGame() {
...
-        logger->log(LOG_FATAL, COMP_ALLEGRO, "Allegro init", allegro_id, OUTC_FAILED);
-        return false;
-    }
-
-    logger->log(LOG_INFO, COMP_ALLEGRO, "Allegro init", allegro_id, OUTC_SUCCESS);
+    // FIXME: eventually, we will want to grab this object in the constructor. But then cGame cannot be a
+    // global anymore, because it needs to be destructed before main exits.
+    _PLInit = std::make_unique<cPlatformLayerInit>("d2tm.cfg");
 
     int r = install_timer();
```
stefanhendriks, line 711: would it make sense to also move all the other `install_` functions from allegro into the platform class? Perhaps in another PR to keep changes small.

stefanhendriks (approved):

stefanhendriks:
I tested this and the game launches, and seems to work fine (shutting down, switching game states, etc).

utils/cPlatformLayerInit.cpp
pbomta: It's a way to cancel construction of an object. The only way (apart from flat-out calling exit ;) )
And you can let the exception catch anywhere in the stack, or nowhere. Some programs have some handler in main or somewhere else on top. Others catch all and return an error code.
In this case, I just let it flow. Interestingly, the console output shows the exception. Exactly what I wanted and without any additional coding. :)

cGame_logic.cpp
pbomta: Yup.
It sets the unique_ptr to nullptr, which deletes the old object it was pointing to.

cGame_logic.cpp
pbomta: I was thinking the same. 
Not sure where the timers should go yet.
Perhaps we can look at how SDL does timers and find some common facade to hide them behind.

Issue #186, opened 2021-05-18
Improve UI voices by mentat


stefanhendriks:
And make mode how chatty it is.

Refering to the sounds like "Construction Complete".

In Dune 2 you also hear `<house> Unit destroyed` and others.

Need to analyze which other sounds are made.

Other examples:

Missile Launched
Base is under attack!

PR #411, opened 2022-01-18
encapsulate all sound and music in cSoundPlayer

pbomta:
- Init and de-init are now done in CSoundPlayer.
- CSoundPlayer reads the audio datafile and owns the data
- Cleaned up some (not all) audio related functions in common.h
- Left mp3 related code behind, as that can be removed in the next PR.

In the next PR, I'd like to move the rest of the audio functions and remove almp3. This PR will be too big, if I do that in one go.

I playtested it and the game works as before.

### DIFF
--- a/building/cItemBuilder.cpp
+++ b/building/cItemBuilder.cpp
@@ -1,6 +1,5 @@
 #include "cItemBuilder.h"
 
-#include "data/gfxaudio.h"
 #include "d2tmc.h"
 #include "gameobjects/structures/cAbstractStructure.h"
 #include "gameobjects/units/cUnit.h"
@@ -11,8 +10,8 @@
 #include "structs.h"
 #include "utils/cLog.h"
 #include "utils/common.h"
+#include "utils/cSoundPlayer.h"
 #include "utils/d2tm_math.h"
-
 
 #include <allegro/keyboard.h>
 
@@ -194,7 +193,7 @@ void cItemBuilder::itemIsDoneBuildingLogic(cBuildingListItem *item) {
         // play voice when placeIt is false`
         if (!item->shouldPlaceIt()) {
             if (player->isHuman()) {
-                play_voice(SOUND_VOICE_01_ATR); // "Construction Complete"
+                game.getSoundPlayer().playVoice(SOUND_VOICE_01_ATR, player->getHouse()); // "Construction Complete"
             }
             item->setPlaceIt(true);
 
--- a/building/cItemBuilder.cpp
+++ b/building/cItemBuilder.cpp
@@ -193,7 +193,7 @@ void cItemBuilder::itemIsDoneBuildingLogic(cBuildingListItem *item) {
         // play voice when placeIt is false`
         if (!item->shouldPlaceIt()) {
             if (player->isHuman()) {
-                game.getSoundPlayer().playVoice(SOUND_VOICE_01_ATR, player->getHouse()); // "Construction Complete"
+                game.playVoice(SOUND_VOICE_01_ATR, player->getHouse()); // "Construction Complete"
             }
             item->setPlaceIt(true);
 
--- a/building/cItemBuilder.cpp
+++ b/building/cItemBuilder.cpp
@@ -433,7 +433,7 @@ void cItemBuilder::deployUnit(cBuildingListItem *item, int buildId) const {
  */
 int cItemBuilder::getFreeSlot() {
 	for (int i = 0; i < MAX_ITEMS; i++) {
-		if (items[i] == NULL) {
+		if (items[i] == nullptr) {
 			return i; // return free slot
 		}
 	}
@@ -517,7 +517,7 @@ cBuildingListItem *cItemBuilder::findBuildingListItemOfSameListAs(cBuildingListI
             return listItem;
         }
 	}
-	return NULL;
+	return nullptr;
 }
 
 /**
--- a/cGame.h
+++ b/cGame.h
@@ -87,7 +87,7 @@ public:
 
 	void setup_players();
 
-    void think_music();
+    void think_audio();
 
 	void think_mentat();
 
@@ -103,9 +103,14 @@ public:
     // Use this instead
 	void setNextStateToTransitionTo(int newState);
 
-	cSoundPlayer& getSoundPlayer() {
-	    return *_soundplayer;
-	}
+    // Game acts as a facade, delegates to sound player
+    void playSound(int sampleId); // Maximum volume
+    void playSound(int sampleId, int vol);
+
+    void playVoice(int sampleId, int house);
+    void playMusic(int sampleId);
+
+    int getMaxVolume();
 
     int getColorFadeSelected(int r, int g, int b) {
         // Fade with all rgb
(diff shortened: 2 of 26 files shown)


utils/cSoundPlayer.cpp
```diff
@@ -5,91 +5,170 @@
...
-	}
+class cSoundData {
+  public:
+    cSoundData() {
+        auto logger = cLogger::getInstance();
+
+        gfxaudio = load_datafile("data/gfxaudio.dat");
+        if (gfxaudio == nullptr) {
+            logger->log(LOG_ERROR, COMP_SOUND, "Initialization", "Could not hook/load datafile: gfxaudio.dat.", OUTC_FAILED);
+            throw 1; // FIXME: real exception
```
pbomta, line 44: Oops, I left that in. I could have sworn I fxed that...

utils/cSoundPlayer.cpp
```diff
@@ -5,91 +5,170 @@
...
+constexpr int kAllegroMaxVolume = 255;
+constexpr int kAllegroPanCenter = 128;
+
+constexpr int kFinishedPLaying = -1;
+constexpr int kNoVoiceAvailable = -1;
+constexpr int kNoVoice = -1;
+constexpr int kNoLoop = 0;
+
+constexpr int kMinNrVoices = 4;
+constexpr int kMaxVolume = 220;
```
pbomta, line 33: This is interesting. Now that more audio stuff is together, we can see there are two different max volumes in the code. 
I think this one can be removed, as all samples are played with either a hardocded volume or the Allegro maximum.
It's only used to set the music volume to 110. (See a FIXME below). I think we can replace this with a max music volume of 110 (or 128?).

cGame.h
```diff
@@ -103,10 +103,8 @@ class cGame : public cScenarioObserver, cInputObserver {
     // Use this instead
 	void setNextStateToTransitionTo(int newState);
 
-	int getMaxVolume() { return iMaxVolume; }
-
-	cSoundPlayer * getSoundPlayer() {
-	    return soundPlayer;
+	cSoundPlayer& getSoundPlayer() {
+	    return *_soundplayer;
```
stefanhendriks, line 107: so this is something I haven't decided yet for naming. I see several approaches, and the google c++ guides also has an opinion about it (but I don't like it).

So these are the possibilities:
`m_<var name>`
`<var name>_`
`_<var name>`

I believe google opts for the second version (appending an underscore).

Personally I don't like either of these options, but you get warnings/errors when you shadow variable names. And being explicit with `this->x = x` is cumbersome perhaps (although fairly common in java).

So what shall we choose?

I lean towards `m_<somevar>` now.

cGame_logic.cpp
```diff
@@ -255,7 +256,7 @@ void cGame::setMissionLost() {
     TIMER_shake = 0;
     mouse->setTile(MOUSE_NORMAL);
 
-    play_voice(SOUND_VOICE_08_ATR);
+    _soundplayer->playVoice(SOUND_VOICE_08_ATR, players[HUMAN].getHouse());
```
stefanhendriks, line 259: I wonder if we should have a `player->playVoice` , which basically does this for us? A `cSoundPlayer` is ofcourse agnostic of which player, the player ID here is used to play the correct sound. In fact, this might be an abstraction to be introduced later? This ties into https://github.com/stefanhendriks/Dune-II---The-Maker/issues/186 so we can chose to defer it for now.

stefanhendriks (changes requested):
Really like how this is going! A few questions, and I think we should not use `getSoundPlayer`. Might be that I try to fix that myself (I have a bit time now).

I really like how we align in how / where to move things to, with some minor details. And it is very nice to see how the code slowly becomes more coherent.

building/cItemBuilder.cpp
```diff
@@ -194,7 +193,7 @@ void cItemBuilder::itemIsDoneBuildingLogic(cBuildingListItem *item) {
         // play voice when placeIt is false`
         if (!item->shouldPlaceIt()) {
             if (player->isHuman()) {
-                play_voice(SOUND_VOICE_01_ATR); // "Construction Complete"
+                game.getSoundPlayer().playVoice(SOUND_VOICE_01_ATR, player->getHouse()); // "Construction Complete"
```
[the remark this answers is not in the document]
stefanhendriks: I've picked this up.

cGame.h
pbomta: > Also interesting, so this function returns the hard pointer, and disregarding the fact it is a `unique_ptr`?

We can tell the caller that they don't have to do a nullptr check. I envision cSoundPlayer to be constructed in the constructor of cGame, then we are certain it's always initialized when we need it. However, cGame is still a global variable, so that is not possible yet...

pbomta: > I lean towards `m_<somevar>` now.

I've used both `m_<something>` and `<something>_` professionally. The m_ is a bit better to see.

utils/cSoundPlayer.cpp
```diff
@@ -5,91 +5,172 @@
...
-				break;
-			}
-		}
+
+	*voice = allocate_voice(sample);
+    assert(*voice != kNoVoiceAvailable);
+
+    voice_set_playmode(*voice, PLAYMODE_PLAY | PLAYMODE_FORWARD); // The default
+    voice_set_volume(*voice, vol);
+    voice_set_pan(*voice, kAllegroPanCenter);
```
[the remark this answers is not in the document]
pbomta: The previous function indeed took a `pan` argument. But it was called only once from common.cpp with a pan of 127. 🤷

cGame.h
stefanhendriks: ok, I suggest to use `m_something` then.

stefanhendriks: > > Also interesting, so this function returns the hard pointer, and disregarding the fact it is a `unique_ptr`?
> 
> We can tell the caller that they don't have to do a nullptr check. I envision cSoundPlayer to be constructed in the constructor of cGame, then we are certain it's always initialized when we need it. However, cGame is still a global variable, so that is not possible yet...

I agree 👍

utils/cSoundPlayer.cpp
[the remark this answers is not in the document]
stefanhendriks: Ah dang it, then I probably broke that functionality earlier before.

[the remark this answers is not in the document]
stefanhendriks: I created an issue to fix this later: https://github.com/stefanhendriks/Dune-II---The-Maker/issues/412

utils/cSoundPlayer.cpp
stefanhendriks: I am unsure about this.  I noticed the max volume returns an Allegro max value, but the intention is to return the max intended volume at some point. So perhaps I made a distinction somewhere sometime for that. 

Regardless, it would make sense to have only one "max" volume  and using the Allegro one would make sense then.

stefanhendriks (approved):
I fixed what I wanted to fix and we discussed everything else and deferred what we wanted. So lgtm now :)

PR #413, opened 2022-01-19
Rename members in cGame class to use m_ prefix

stefanhendriks:
This PR makes sure all member fields in `cGame` use the `m_<camelCaseVariableName>` consistently.

I also went over some functions and grouped them if it made sense, or renamed them where applicable. Making sure all functions are also using the `camelCase()` convention. There are only a few exceptions for now with `state_` or `think[fast/slow]_` prefixes.


pbomta (approved):
Nice

PR #415, opened 2022-01-23
remove almp3

pbomta:
also moves rest of sound related functions from common to cGame

Closes: #388


cGame.h
```diff
@@ -79,7 +78,7 @@ class cGame : public cScenarioObserver, cInputObserver {
 
 	int m_pathsCreated;
 
-    int m_musicVolume;              // volume of the mp3 / midi
+    int m_musicVolume;              // volume of the music
```
pbomta, line 81: This data member is not used. Shall we leave it in for now? I believe the plan is to have volume settings in the menu at some point.

stefanhendriks:
I feel sad to see MP3 support go, but I know why it happens. I have fond memories of having MP3 files with more adlib or MT32 sounds of the Dune 2 music. This will return once we can migrate to SDL or something. All for the better.

Nice improvements otherwise!

stefanhendriks:
(btw, MP3 support was broken anyway, so not that sad ;-) )

cGame.h
stefanhendriks: 🤔 , I guess we can keep it. Once we have a proper configuration object we can move it to there.

stefanhendriks (approved):

PR #416, opened 2022-01-23
replace definition of pi with standard library one

pbomta:
also some cleanup of math functions and includes


stefanhendriks:
Ohh nice, touched really old code I see. And good to see the generic functions be replaced by the `map*` ones, as it should be 👍

stefanhendriks (approved):

PR #422, opened 2022-01-26
Update alfont to 2.0.9 and build it using CMake with the rest of the build

pbomta:
* ALFONT_DLL is now set in the CMakeLists file (step towards removing allegroh.h)
* Interesting: the resulting binary (d2tm.exe) is smaller.

Playtested it and it works!


stefanhendriks:
I wonder, would it be better to statically compile alfont within d2tm? (remove the need of a dll?). I believe we can do that now right?

stefanhendriks:
I noticed, when opening alfont source file is in WIN-1252 according to my CLion. I believe all other is UTF8. Is it wise to convert alfont as well to UTF8?

Also, I have some compiler warnings & cppcheck complains about potential memory leaks in `alfont.c`, do we also need to fix those? Or can we suppress them?

Since we're going to migrate to another library , I don't think it is worth the effort of fixing alfont though.

All in all awesome that this is done. I can't imagine doing it myself as quickly as you did 👍

stefanhendriks:
These are my `cppcheck` errors btw: (line numbers at the end)

```
Cppcheck: (error) memleak: Memory leak: s_pointer (:844)
Cppcheck: (error) memleak: Memory leak: s_pointer (:2007)
Cppcheck: (error) uninitvar: Uninitialized variable: lpszWA (:3809)
Cppcheck: (error) uninitvar: Uninitialized variable: lpszWA (:4125)
```

stefanhendriks:
Other than the above, it works also on my end :)

stefanhendriks (approved):

pbomta:
> I wonder, would it be better to statically compile alfont within d2tm? (remove the need of a dll?). I believe we can do that now right?

Certainly. I'll make it so.

Issue #188, opened 2021-05-19
Only allow cheat & debug key's when `-debug` switch is given


stefanhendriks:
Enable them with startup flag.

PR #423, opened 2022-01-26
Debug mode keys when `debug` mode is on

stefanhendriks:
for - #188

This also makes most code that uses `key[<SCANCODE>]` react on the keyboard events.  The exceptions are on units or other places where we remove the code or defer to later. 

To test: start game without `-debug` flag and you should not be able to use debug keys (`TAB`- etc). 

- most obvious things player noticed where changing player (`TAB-<nr>`)


building/cItemBuilder.cpp
```diff
@@ -758,3 +759,32 @@ void cItemBuilder::removeItemsByBuildId(eBuildType buildType, int buildId) {
...
+            onKeyPressed(event);
+            break;
+        default:
+            break;
+    }
+}
+
+void cItemBuilder::onKeyHold(const cKeyboardEvent &event) {
+    if (event.hasEitherKey(KEY_LSHIFT, KEY_RSHIFT)) {
+        m_buildItemMultiplierEnabled = true;
```
pbomta, line 781: Nice. Can this class now get rid of the allegro keyboard include?

drawers/cBuildingListDrawer.h
```diff
@@ -1,12 +1,17 @@
...
 
-class cBuildingListDrawer {
+#include "observers/cInputObserver.h"
+
+class cBuildingListDrawer : cInputObserver {
 
 public:
-	cBuildingListDrawer(cPlayer *thePlayer);
-	~cBuildingListDrawer();
+	explicit cBuildingListDrawer(cPlayer *player);
```
pbomta, line 8: 👍

tools/LogViewer/nbproject/build-impl.xml
```diff
@@ -491,7 +491,7 @@ is divided into following sections:
     </target>
     <!--
                 =================
-                DEBUGGING SECTION
+                game.isDebugMode() SECTION
```
pbomta, line 494: this was probably not meant to be renamed :) also other .xml files

utils/cLog.h
```diff
@@ -62,5 +64,7 @@ class cLogger {
 
       long startTime; // start time of logging in miliseconds
 
+      bool m_debugMode;
```
pbomta, line 67: Eventually, we could let the debug mode be dependent on the log level, but that's not easy to do half-half. We'll need to decide that and change all the calls at once, I expect.

pbomta (approved):

controls/cKeyboardEvent.h
```diff
@@ -32,10 +32,9 @@ class cKeyboardEvent {
         str.append(partOne);
         str.append("[");
         for (auto key : event.keys) {
-            str.append(std::to_string(key));
-            str.append("=");
-            str.append(std::to_string(scancode_to_ascii(key)));
-            str.append(" ");
+            char msg[10];
+            sprintf(msg, "%s,", scancode_to_name(key));
```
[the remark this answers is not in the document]
stefanhendriks: I don't know why this didn't work ( I tried `std.append` first). I'll retry later.

tools/LogViewer/nbproject/build-impl.xml
stefanhendriks: 🤦‍♂️

utils/cLog.h
stefanhendriks: Hmm. Well, in the end `debugMode` basically enables some 'flags' to allow us to see (visually, like drawing on screen) different things. It *also* has impact on logging atm.

So those things need to be untangled:
- log levels (like you said)
- debug/drawing information per feature perhaps? (with some kind of variables)

For some other time for sure :)

controls/cKeyboardEvent.h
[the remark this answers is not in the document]
stefanhendriks: fixed

PR #424, opened 2022-01-27
Update Allegro to 4.4 and build it using CMake

pbomta:
Could not build Allegro statically, it is still a DLL named alleg44.dll. Apart from the standard Windows dependencies, it has no other dependencies. So no more copying of dlls from mingw.


stefanhendriks:
Again, I am wowed by this. This has gone so fast, and it works flawlessly. I just pull, build, and run. It works. Awesome. Also within CLion, it just works:
```
Consolidate compiler generated dependencies of target fmt
[  1%] Built target fmt
Consolidate compiler generated dependencies of target allegro
[ 45%] Built target allegro
Consolidate compiler generated dependencies of target alfont
[ 55%] Built target alfont
Consolidate compiler generated dependencies of target d2tm
[100%] Built target d2tm

```

This is so satisfying :) 

Ok, so aside all the drooling, here are a few nitpicks:
- there is `create binary` bat file, that could probably depend on the `rebuildwin.bat` thing (and possibly be even made into a single script able to run on all OS's? anyway, I digress). 
- I know allegro can be statically compiled; but is this something not supported or built in with CMake in this version? (see also: https://liballeg.org/stabledocs/en/build/cmake.html)

stefanhendriks:
Lastly, maybe we want to make our CI a bit smarter by caching the results of the dependencies? Or should we rebuild those everytime? Seems like a waste of time?

stefanhendriks (approved):

stefanhendriks:
Once this is merged #58 should become possible! 🎉

pbomta:
> * there is `create binary` bat file, that could probably depend on the `rebuildwin.bat` thing (and possibly be even made into a single script able to run on all OS's?

I would expect mingw supports shell scripts. If we make a regular shell script (so no bash script), then that can run on Linux, in MSYS2 and on a Mac.

pbomta:
> * I know allegro can be statically compiled; but is this something not supported or built in with CMake in this version?

CMake is not the problem. I think there is a mistake in the CMakeLists.txt that comes with Allegro 4.4. If I set it to build a static library, I get a large list of unresolved functions. I looked into it for a while, but eventually gave up.

pbomta:
> Again, I am wowed by this. This has gone so fast, and it works flawlessly.

Thanks a lot! I'm also happy to get it to work.

pbomta:
> Lastly, maybe we want to make our CI a bit smarter by caching the results of the dependencies?

Maybe, but Allegro is C code that builds really fast. So I wouldn't bother.

PR #425, opened 2022-01-28
Include allegro where used

pbomta:
Removed allegroh.h and included allegro where it is needed (many places still...)
Enabled link time optimization because the binary size increased with a megabyte otherwise.


stefanhendriks:
Good catch for compiling release mode. Perhaps let CI always build in release mode? (or both?)

You're on a roll btw! nice improvements.

stefanhendriks (approved):

pbomta:
> Perhaps let CI always build in release mode? (or both?)

I think the default is fine. When developing, we can keep using the default (Debug). Only when releasing, we can use Release mode. In a more professional setting, we would probably build both and even package a release on each commit, but that's perhaps a bit overkill at the moment. :)

PR #430, opened 2022-02-03
modernize cSeedMapGenerator a bit and include it only where used

pbomta:
Some cleanup. Returning the seedmap from the generator by value, as NRVO (named return value optimization) will construct it in-place and no copy will be done. In case that fails, the seedmap contains only one vector, which supports move semantics, so the data buffer is not copied if the original object is not used anymore.

### DIFF
--- a/ini.cpp
+++ b/ini.cpp
@@ -13,6 +13,7 @@
 #include "include/d2tmh.h"
 
 #include "utils/cLog.h"
+#include "utils/cSeedMapGenerator.h"
 
 #include <allegro.h>
 #include <fmt/core.h>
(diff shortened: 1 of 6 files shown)


stefanhendriks:
I'd need to test this a bit in campaign mode but it looks ok to me :)

stefanhendriks:
I'll test this now

stefanhendriks:
I start an `Atreides` campaign , and I immediately get a crash (SIGSEV):

image

Not sure why though.

stefanhendriks:
I unfolded `map` in debug view:

image

Could it be that it tries to insert at index `14` within the vector while there is only `0` and `1` and you need to have them in order? (can't be having gaps I suppose, as it is more like an array right?)

stefanhendriks:
When retrying I get a SIGSEV on a totally different place. 

I'm not sure what happened, this would require some backtracking and figuring out when it went wrong (git is our friend).

stefanhendriks (changes requested):
Please investigate why it crashes :)

stefanhendriks:
FYI, I tested on `master` and that works

pbomta:
> Please investigate why it crashes :)

Ah, I just saw this. Thanks for testing! I'll take a look.
When I tested it, it worked. Good that we're more than one person.

pbomta:
> Could it be that it tries to insert at index `14` within the vector while there is only `0` and `1`

Oh, no! Of course. I ran into a tricky corner of C++ that several blogposts warn you for! 
See here: https://hackingcpp.com/cpp/std/vector.html expand "Make New Vectors". There the case is explained.

pbomta:
I tested it only in skirmish mode, but now I see that doesn't even use the seedmap.

stefanhendriks:
> 

true. Skirmish mode has a different map generator.

stefanhendriks:
> > Could it be that it tries to insert at index `14` within the vector while there is only `0` and `1`
> 
> Oh, no! Of course. I ran into a tricky corner of C++ that several blogposts warn you for! See here: https://hackingcpp.com/cpp/std/vector.html expand "Make New Vectors". There the case is explained.

haha, nice. It is a subtle difference but has great impact ;-)

PR #431, opened 2022-02-03
removes some redundant functions from common.h

pbomta:
also some header cleanup

67 fewer lines of code == fewer bugs ;)

### DIFF
--- a/controls/mousestates/cMousePlaceState.cpp
+++ b/controls/mousestates/cMousePlaceState.cpp
@@ -82,8 +82,8 @@ bool cMousePlaceState::mayPlaceIt(cBuildingListItem *itemToPlace, int mouseCell)
     int iEndY = iCellY + SCANWIDTH + cellHeight;
 
     // Fix up the boundaries
-    map.fixCoordinatesToBeWithinMap(iStartX, iStartY);
-    map.fixCoordinatesToBeWithinMap(iEndX, iEndY);
+    cPoint::split(iStartX, iStartY) = map.fixCoordinatesToBeWithinMap(iStartX, iStartY);
+    cPoint::split(iEndX, iEndY) = map.fixCoordinatesToBeWithinMap(iEndX, iEndY);
 
     // Determine if structure to be placed is within build distance
     for (int iX = iStartX; iX < iEndX; iX++) {
--- a/controls/mousestates/cMousePlaceState.cpp
+++ b/controls/mousestates/cMousePlaceState.cpp
@@ -82,8 +82,8 @@ bool cMousePlaceState::mayPlaceIt(cBuildingListItem *itemToPlace, int mouseCell)
     int iEndY = iCellY + SCANWIDTH + cellHeight;
 
     // Fix up the boundaries
-    FIX_POS(iStartX, iStartY);
-    FIX_POS(iEndX, iEndY);
+    map.fixCoordinatesToBeWithinMap(iStartX, iStartY);
+    map.fixCoordinatesToBeWithinMap(iEndX, iEndY);
 
     // Determine if structure to be placed is within build distance
     for (int iX = iStartX; iX < iEndX; iX++) {
--- a/map/cMap.cpp
+++ b/map/cMap.cpp
@@ -1139,16 +1139,13 @@ bool cMap::isAtMapBoundaries(int cell) {
     return false;
 }
 
-void cMap::fixCoordinatesToBeWithinPlayableMap(int &x, int &y) const {
-    x = std::clamp(x, 1, getWidth() - 2);
-    y = std::clamp(y, 1, getHeight() - 2);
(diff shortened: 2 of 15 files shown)


controls/mousestates/cMousePlaceState.cpp
```diff
@@ -82,8 +82,8 @@ bool cMousePlaceState::mayPlaceIt(cBuildingListItem *itemToPlace, int mouseCell)
     int iEndY = iCellY + SCANWIDTH + cellHeight;
 
     // Fix up the boundaries
-    FIX_POS(iStartX, iStartY);
-    FIX_POS(iEndX, iEndY);
+    map.fixCoordinatesToBeWithinMap(iStartX, iStartY);
```
stefanhendriks, line 85: tbh, I really hate this function. As it changes the values using out vars. I'd rather have some function that returns corrected values (even though it is a bit more verbose). Anyway, for now this is 👌

stefanhendriks: with "this function" i meant "FIX_POS()" ;-)

map/cMap.cpp
```diff
@@ -1139,10 +1139,17 @@ bool cMap::isAtMapBoundaries(int cell) {
...
 }
 
-void cMap::fixCoordinatesToBeWithinPlayableMap(int &x, int &y) {
-    FIX_BORDER_POS(x, y);
+void cMap::fixCoordinatesToBeWithinPlayableMap(int &x, int &y) const {
+    x = std::clamp(x, 1, getWidth() - 2);
+    y = std::clamp(y, 1, getHeight() - 2);
 }
 
+void cMap::fixCoordinatesToBeWithinMap(int &x, int&y) const {
```
stefanhendriks, line 1147: `std::clamp` is quite nice :)

Even better would be, at some point, to have it accept a `cPoint` and return a `cPoint` or something similar. For now we use the same mechanism, so thats fine.

stefanhendriks:
Happy to see those generic functions go and have them moved to where they belong. My hands itch to have some unit tests for such functions, but I don't know how to do that (yet).

Although I hated my approach of `FIX_POS` , and we keep the way it works (which is fine), at some point I probably would refactor it to use `cPoint`'s or something.

stefanhendriks (approved):

controls/mousestates/cMousePlaceState.cpp
pbomta: Output parameters are indeed frowned upon nowadays. But with C++17 and the "structured bindings" feature, we can return multiple values. I'll see what I can do in this PR.

map/cMapEditor.cpp
```diff
@@ -100,7 +100,7 @@ void cMapEditor::createRandomField(int cell, int terrainType, int size) {
                 break;
         }
 
-        map.fixCoordinatesToBeWithinMap(x, y);
+        cPoint::split(x, y) = map.fixCoordinatesToBeWithinMap(x, y);
```
stefanhendriks-felyx, line 103: ok let me try to unpack this:

- this makes `map.fixCoordinatesToBeWithinMap` return a `split` struct (without needing to specifiy in the function body ? ( as I see `{std::clamp(x, 1, getWidth() - 2), std::clamp(y, 1, getHeight() - 2)}`). This is a return value (not by reference)
- and `cPoint::split` basically holds 2 values by reference.

This also means a `cPoint::split{}` is something different than a `cPoint` correct? 

If so, what is the benefit of using a `split` struct over returning a `cPoint` itself?

pbomta: Ah, no, it works a bit differently.
TBH, I forgot to add an explanation for it. It's not easy C++, as it uses a temporary object with references to the target variables to put the cPoint's coordinates in. I've added an explanation. Does that explain enough?
I like to explain some details in a project like this, as we can't (and shouldn't) assume any minimum skill level of contributors, IMHO.

pbomta: Oh, and the benefit is that we can write something like:
```
cPoint::split(x, y) = getPoint()
```
 instead of:
```
auto p = getPoint();
x = p.x;
y = p.y;
```
I do think that in many cases, if we add overloads (or replace) functions that take a coordinate pair to accept cPoint, the split helper class is no longer needed.

utils/cPoint.h
```diff
@@ -16,8 +16,16 @@ class cPoint {
 
     bool isWithinRectangle(const cRectangle *pRectangle) const;
 
+    /**
+     * Helper class to split a point that is returned from a function into its two coordinates.
+     * It stores two references to the coordinate variables. The copy operator copies the coordinates
+     * from the given point into the variables pointed to by the references.
+     * Example:
+     *   int x, y;
+     *   cPoint::split(x, y) = getPoint();
```
stefanhendriks, line 25: aha this makes sense. So this is a convenient way to get the two raw x,y coordinates without `.x` or `.y` 'ing this. Nice.

PR #432, opened 2022-02-03
move cTimeManager inside cGame

pbomta:
Another global variable gone.
I wonder if we should keep the time manager. Perhaps in a later refactoring, the handling of the allegro timers can be done in the main game loop.


stefanhendriks-felyx:
Nice. Even better, the way we deal with time should be done without any Allegro stuff but we should deal with time delta's between frames and such.

But, before that, this is a nice intermediate step !

stefanhendriks-felyx (approved):

Issue #434, opened 2022-02-04
Old mouse cursor still drawn when going to options menu


pbomta:
See? Two mouse cursors. :) 
![d2tm-mouse-cursor-shown](https://user-images.githubusercontent.com/13233624/152533257-f5314c47-c16a-40c7-b6f9-495148f80851.png)

stefanhendriks:
I also noticed this.

Btw, FYI: when I test and it is in between versions I always create a ticket and assign the `playtesting` label.

This is so that when I create release notes this won't show up. A 'bug' is something that was found in the previous released version(s).

pbomta:
> Btw, FYI: when I test and it is in between versions I always create a ticket and assign the `playtesting` label.

Oh, yes, that makes sense. I was already wondering how you handled this.

PR #435, opened 2022-02-04
redraw the screen as a background of the options menu without mouse cursor

pbomta:
Closes: #434


stefanhendriks:
Thats really a nice fix btw, I didn't realize it was becoming this easy to fix (reason why I deferred this). Nice 💪

PR #436, opened 2022-02-05
remove a duplicated rectangle in cGuiButton

pbomta:
Ran cppheck with warnings and found (among many) this one: cGuiButton had a rect member and its parent cGuiObject too. Fixed it and applied the new naming style and include style.


stefanhendriks:
Nice improvements! lgtm

PR #439, opened 2022-02-06
include cleanup of cDrawManager and all drawers

pbomta:
also construct draw manager member objects with the draw manager, instead of on the heap somewhere.


drawers/cMouseDrawer.cpp
```diff
@@ -110,14 +113,13 @@ void cMouseDrawer::drawToolTip() {
 		int structureType = theStructure->getType();
 		if (structureType == WINDTRAP) {
 			cWindTrap * windTrap = dynamic_cast<cWindTrap*>(theStructure);
+            assert(windTrap);
```
stefanhendriks, line 116: do we need this assert? (was it used for testing perhaps?)

drawers/cMouseDrawer.cpp
```diff
@@ -167,21 +169,19 @@ void cMouseDrawer::drawToolTipBackground() {
...
-	assert(textWriter);
 	if (theStructure->belongsTo(player)) {
-		textWriter->writeWithOneInteger("Sight : %d", theStructure->getSight());
-		textWriter->writeWithOneInteger("Range : %d", theStructure->getRange());
+		textWriter.writeWithOneInteger("Sight : %d", theStructure->getSight());
+		textWriter.writeWithOneInteger("Range : %d", theStructure->getRange());
 	} else {
-		textWriter->write("Sight : Unknown");
-		textWriter->write("Range : Unknown");
+		textWriter.write("Range : Unknown");
```
stefanhendriks, line 178: Range/Side order got swapped here. (compared to diff & compared to other block in `if`

drawers/cMouseDrawer.cpp
```diff
@@ -190,58 +190,58 @@ void cMouseDrawer::drawToolTipGeneralInformation(cAbstractStructure * theStructu
 	} else {
 		sprintf(description, "%s", structureType.name);
 	}
-	textWriter->write(description, makecol(255, 255, 0));
-	textWriter->writeWithTwoIntegers("Hitpoints : %d/%d", theStructure->getHitPoints(), theStructure->getMaxHP());
-	textWriter->writeWithOneInteger("Armor : %d", theStructure->getArmor());
-	textWriter->writeWithOneInteger("Protected : %d%%", (100-theStructure->getPercentageNotPaved()));
+	textWriter.write(description, makecol(255, 255, 0));
+	textWriter.writeWithTwoIntegers("Hitpoints : %d/%d", theStructure->getHitPoints(), theStructure->getMaxHP());
```
stefanhendriks, line 194: we can - eventually - get rid of `writeWith...` by combining `fmt::format` with `write()`. Not for now though :)

drawers/cMouseDrawer.cpp
```diff
@@ -190,58 +190,58 @@ void cMouseDrawer::drawToolTipGeneralInformation(cAbstractStructure * theStructu
...
-	assert(textWriter);
-
+	
 	// TODO / IDEA --> Perhaps some 'spy on enemy intel' upgrade should be available for
 	// house Ordos, so you can actually check on the enemy spice etc.
 	if (theStructure->getOwner() == HUMAN) {
 		int spiceCapacityOfStructure = 0;
 		if (theStructure->getType() == REFINERY) {
 			cRefinery * refinery = dynamic_cast<cRefinery*>(theStructure);
+            assert(refinery);
```
stefanhendriks, line 226: Additional asserts added? 

I do wonder if we can't solve this prettier. Instead of relying on `type` (REFINERY for instance), we should perhaps have some other way to get this information. Not sure how. My gut feel tells me that a sort of `composit` would make it possible to add some behavior to some structures which have a `capacity` of some sort. But I didn't spend too many brain cycles on this. Do you have any ideas about this?

drawers/cUnitDrawer.cpp
```diff
@@ -1,27 +1,22 @@
...
 cUnitDrawer::cUnitDrawer() {
 }
 
-cUnitDrawer *cUnitDrawer::getInstance() {
-	if (instance == NULL) {
-		instance = new cUnitDrawer();
-	}
-
-	return instance;
+const cUnitDrawer& cUnitDrawer::getInstance() {
```
stefanhendriks, line 6: afaik we don't use this drawer?

stefanhendriks (approved):
looks like a nice improvement to me, I've got a few minor questions.

I also want to test this a bit before merging. Seems to have quite some impact, but overall looking nice. Good to see more and more consistency in the code 🚀

drawers/cBuildingListDrawer.cpp
```diff
@@ -1,19 +1,18 @@
...
+#include "cAllegroDrawer.h"
+#include "d2tmc.h"
+#include "data/gfxinter.h"
+#include "player/cPlayer.h"
 
 #include <allegro.h>
 
 cBuildingListDrawer::cBuildingListDrawer(cPlayer *player) :
-    m_textDrawer(new cTextDrawer(game_font)),
+    m_textDrawer(game_font),
```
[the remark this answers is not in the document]
pbomta: Uhm, not really. This constructs `m_textDrawer` in-place inside `cBuildingListDrawer` so-to-speak. It's object memory is layout inside the object memory of `cBuildingListDrawer`. What you typed uses the copy constructor (unless that is optimized out - never know what the optimizer can do nowadays). So that would construct `cTextDrawer` on the stack and then copy (or move if it can) it into `cBuildingListDrawer`.

TL;DR: this is a way to use data member with non-default constructors.

drawers/cUnitDrawer.cpp
stefanhendriks-felyx: deleting is ok :)

drawers/cMouseDrawer.cpp
stefanhendriks: Yeah I can imagine. It feels awkward because we only have 1 or 2 buildings which will actually implement this :/

PR #440, opened 2022-02-07
fix several cppcheck warnings in cLog and in gameobjects

pbomta:
Some unused variables an uninitialized things.
I refactored cGameObject to use cPoint, as we discussed when talking about cMap. I think that new classes should use a point and not separate x & y variables. But you may want to delete cGameObject for now? Let me know.


gameobjects/cGameObject.cpp
```diff
@@ -1,17 +1,7 @@
-/*
- * cGameObject.cpp
- *
- *  Created on: 7-nov-2010
- *      Author: Stefan
- */
-
 #include "cGameObject.h"
```
stefanhendriks, line 1: afaik we don't use `cGameObject`. it was introduced with an idea but never pursued this. We could perhaps delete this and re-introduce when really needed later?

stefanhendriks (approved):
Lgtm overall. Nice improvements!

stefanhendriks:
as for `I think that new classes should use a point and not separate x & y variables`

You are right. We should get rid of separate `x,y` as much as possible.

In fact, I'd want to get rid of `cell` as well if possible/handy. In the end I think structures, units, etc are all positioned on 'absolute coordinates' on a map. These absolute coordinates map with pixels btw. (which is fine I think for now)

gameobjects/cGameObject.cpp
pbomta: Sounds good. So far, there is the split between an abstract structure and abstract unit, which works fine.

PR #445, opened 2022-02-13
add Linux build fix for alfont

pbomta:
Not tested on Linux yet, as I don't have access to such a system right now.
But this is the CMake trick that can make it build on both MinGW on Windows and on Linux.

### DIFF
--- a/dependencies/alfont/CMakeLists.txt
+++ b/dependencies/alfont/CMakeLists.txt
@@ -48,9 +48,18 @@ target_link_libraries(alfont
     allegro
 )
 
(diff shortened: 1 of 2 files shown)


stefanhendriks-felyx:
Looks good to me. I don't have a linux system myself. We could perhaps test this on a different github action? 🤷

pbomta:
> We could perhaps test this on a different github action? 🤷
The rest doesn't build yet...

stefanhendriks:
👍 ok lets defer GH action to later. Lets merge this.

PR #448, opened 2022-02-14
construct map editor only when needed

pbomta:
Easy cleanup of a global object.
The random map generator is next (depends on this PR).


include/d2tmc.h
```diff
@@ -30,7 +30,6 @@ class cAllegroDrawer;
 class cBullet;
 class cDrawManager;
 class cMapCamera;
-class cMapEditor;
```
stefanhendriks, line 33: 🎉

include/d2tmc.h
```diff
@@ -50,7 +49,6 @@ extern int		iRest;	// rest value
 extern cGame		  game;
 
 extern cMap           map;
-extern cMapEditor	  mapEditor;
```
stefanhendriks, line 53: 🎉

stefanhendriks (approved):
Good to see it go like that. Nice!

stefanhendriks:
I don't see changes that I think should be tested (it either works or it doesn't). So feel free to merge.

PR #449, opened 2022-02-15
include draw & interaction managers only where needed

pbomta:


stefanhendriks:
lgtm!

ini.cpp
```diff
@@ -13,6 +13,7 @@
 #include "include/d2tmh.h"
 
 #include "map/cMapEditor.h"
+#include "managers/cDrawManager.h" // TODO: an ini file reader should not depend on drawing code
```
stefanhendriks, line 16: this is weird indeed

stefanhendriks (approved):

PR #451, opened 2022-02-15
includes ini component only where needed and removes utilsh header

pbomta:


stefanhendriks:
Good to see another "global" header file go away 👍

Issue #450, opened 2022-02-15
Crash on exit  -  Linux OS


Akatsuki57:
Game crash when i click on exit.

![Capture d’écran_2022-02-15_12-55-41](https://user-images.githubusercontent.com/99535165/154057650-258cdefb-f170-45f0-91e0-41a3b4356fb9.png)

gdb tell  me it's came from alfont

```c++
    // Destroy font of Allegro FONT library
    alfont_destroy_font(game_font);
    alfont_destroy_font(bene_font);
```

I change this in 

```c++
    // Destroy font of Allegro FONT library
    // alfont_destroy_font(game_font);
    // alfont_destroy_font(bene_font);
```

No game crash

stefanhendriks:
the `alfont_destroy_font` calls should be done at the proper moment. Needs investigation.

Also which branch do you use right now? As the linux compatibility branch still needs to be merged into `master`.

PR #459, opened 2022-02-18
Os build

Akatsuki57:
To distinguish linux and windows

### DIFF
--- a/cGame_logic.cpp
+++ b/cGame_logic.cpp
@@ -39,6 +39,13 @@ constexpr auto kMaxAlpha = 255;
 
 }
 
+#ifdef D2TM_LINUX	//run in LINUX
+    #define OS_PLATFORM "Linux OS"
+#endif
+#ifdef D2TM_DEFAULTS	//run in Defaults
+    #define OS_PLATFORM "Default OS"
+#endif
+
 cGame::cGame() : m_timeManager(*this) {
     memset(m_states, 0, sizeof(cGameState *));
 
--- a/cGame_logic.cpp
+++ b/cGame_logic.cpp
@@ -40,9 +40,9 @@ constexpr auto kMaxAlpha = 255;
 }
 
 #ifdef D2TM_LINUX	//run in LINUX
-    #define OS_PLATFORM "Linux OS"
+    constexpr auto OS_PLATFORM = "Linux OS" ;
 #else //#ifdef D2TM_DEFAULTS	//run in Defaults
-    #define OS_PLATFORM "Default OS"
+    constexpr auto OS_PLATFORM = "Default OS"; 
 #endif
 
 cGame::cGame() : m_timeManager(*this) {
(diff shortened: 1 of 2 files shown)


cGame_logic.cpp
```diff
@@ -645,9 +651,10 @@ void cGame::shutdown() {
     }
 
     // Destroy font of Allegro FONT library
+#ifdef D2TM_DEFAULTS
```
stefanhendriks, line 654: I know why you do this (as you said it breaks on linux). But I do wonder why it breaks there. Is it because of alfont itself? Or perhaps the resources it tries to destroy (`game_font` and `bene_font`) are somehow no longer valid here?

Perhaps if you can find out the reason why this is needed, we can add a comment for that. 

In the end, when migrating to SDL2 this will be removed anyway :)

Akatsuki57: I investigate a little more. `alfont_destroy_font` crash without explication. Normally, I would have been more conscientious, memory leaks are bad. But since alfont is old, deprecated, not well documented, and we're going to replace it, I won't spend any more time on it.

pbomta: Perhaps we don't need to destrouy the font at all?
We don't destroy small_font, but only these two.
The next line is, alfont_exit. I hope that would destroy fonts. I couldn't find any documentation...

I would suggest commenting out the delete_font lines (and adding a commented out one about small_font as well) with a comment above explaining that this crashes and that we think it is not needed anyway.

Then we don't need that OS define. Ideally, we would have no OS specific code in the d2tm code at all.

stefanhendriks: good point, if we don't destroy the font at all we can leave out the `#ifdef` part. And then we have best of both worlds.

cGame_logic.cpp
```diff
@@ -694,7 +701,7 @@ bool cGame::setupGame() {
 
 	logger->logHeader("Version information");
 	logger->log(LOG_INFO, COMP_VERSION, "Initializing",
-              fmt::format("Version {}, Compiled at {} , {}", game.m_version, __DATE__, __TIME__));
+              fmt::format("Version {}, Compiled at {} , {} on {}", game.m_version, __DATE__, __TIME__, OS_PLATFORM));
```
stefanhendriks, line 704: I think this detail is pretty neat :)

stefanhendriks (changes requested):
Please remove the `ifdef` part for destroying fonts, lets not destroy them so it won't crash on Linux. As you said - once we migrate to SDL this won't be a problem anymore :)

cGame_logic.cpp
Akatsuki57: Yes but it's not very practical since you have to be independent of the OS as @pbomta  says. And i create a dependency. 

I saw that the SDL2 already had the function to know where we are.
https://wiki.libsdl.org/SDL_GetPlatform

Akatsuki57:
I then propose to simply delete this branch and correcting bug #450 apart

cGame_logic.cpp
stefanhendriks: I'll leave it up to you how to deal with this. It is not a mandatory feature, but I do think a nice touch. If you think we should not do it, feel free to remove it. (although I think the dependency here is less of a problem compared to the `#ifdef` above?)

stefanhendriks:
@Akatsuki57 ok!

PR #460, opened 2022-02-18
Fullscreen can be set by game.ini

Akatsuki57:
Just the possibility to set/unset fullscreen mode in game.ini with other parameters like width and height.
All was already done before

### DIFF
--- a/game.ini
+++ b/game.ini
@@ -7,6 +7,7 @@
 #Uncomment to hard set resolution, keep commented to use auto-detection
 ScreenWidth=800
 ScreenHeight=600
+#FullScreen=false
 CameraBorderOrKeyMoveSpeed=0.5
 CameraDragMoveSpeed=0.5
 CameraEdgeMove=true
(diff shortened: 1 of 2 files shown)


pbomta (approved):

stefanhendriks:
So I remember having disabled this feature a while back. And the rationale was as follows:

You can boot `d2tm` with `-windowed`. But if you also have set `windowed=false` in the `ini` file, the game would still run in fullscreen mode. 

However:
I would expect to have program arguments supersede anything in any config file. Hence, running with `-windowed` flag should *always* run in windowed mode. Regardless of any config file setting.

Unfortunately, with the way things are set-up this won't work.

So, ideally we can store somewhere if d2tm should be running in full-screen mode or not (via `game.ini` or something else). But since it is not possible to supersede this , I'd say we have to fix that first before we can merge this.

To properly fix this we would need to redo the initialization part of the game. 

In abstract sense I think this should become the picture:

- read startup params
- read config file
- merge two together to determine 'configuration state'
- initialize game with this 'configuration state'
- proceed...

I have created a ticket (https://github.com/stefanhendriks/Dune-II---The-Maker/issues/466) for this. Lets skip this PR for now, and do this later properly once we fixed #466 :)

stefanhendriks (changes requested):
Lets do #466 first 👍

Akatsuki57:
integrated in the pull request #488

PR #462, opened 2022-02-19
proper includes for gameobjects

pbomta:


gameobjects/units/cUnitUtils.h
```diff
@@ -10,23 +10,15 @@
...
+class cUnitUtils {
+  protected:
+  	cUnitUtils();
 
-	int findUnit(int type, int iPlayerId); // find the first unit of type belonging to player Id.
-	int findUnit(int type, int iPlayerId, int iIgnoreUnitId);
+  public:
+	  static cUnitUtils& getInstance();
 
+	  int findUnit(int type, int iPlayerId); // find the first unit of type belonging to player Id.
```
stefanhendriks, line 22: Is this even used? If not, we can delete it.

I believe some functions have been added to `player` for finding units and such (which are used)

PR #463, opened 2022-02-19
Fmt control

Akatsuki57:
More fmt::format instead of sprintf
cLog use now fsteam

### DIFF
--- a/cGame_logic.cpp
+++ b/cGame_logic.cpp
@@ -1623,19 +1623,10 @@ void cGame::drawCombatMouse() {
 }
 
 void cGame::saveBmpScreenToDisk() {
-    char filename[25];
 
-    if (m_screenshot < 10) {
-        sprintf(filename, "%dx%d_000%d.bmp", m_screenX, m_screenY, m_screenshot);
-    } else if (m_screenshot < 100) {
-        sprintf(filename, "%dx%d_00%d.bmp", m_screenX, m_screenY, m_screenshot);
-    } else if (m_screenshot < 1000) {
-        sprintf(filename, "%dx%d_0%d.bmp", m_screenX, m_screenY, m_screenshot);
-    } else {
-        sprintf(filename, "%dx%d_%d.bmp", m_screenX, m_screenY, m_screenshot);
-    }
+    std::string filename = fmt::format("{}x{}_{:0^4}.bmp", m_screenX, m_screenY, m_screenshot);
 
-    save_bmp(filename, bmp_screen, general_palette);
+    save_bmp(filename.c_str(), bmp_screen, general_palette);
 
     // shows a message in-game, would be even better to have this 'globally' (not depending on state), kind of like
     // a Quake console perhaps?
--- a/cGame_logic.cpp
+++ b/cGame_logic.cpp
@@ -1624,7 +1624,7 @@ void cGame::drawCombatMouse() {
 
 void cGame::saveBmpScreenToDisk() {
 
-    std::string filename = fmt::format("{}x{}_{:0^4}.bmp", m_screenX, m_screenY, m_screenshot);
+    std::string filename = fmt::format("{}x{}_{:0>4}.bmp", m_screenX, m_screenY, m_screenshot);
 
     save_bmp(filename.c_str(), bmp_screen, general_palette);
 
--- a/drawers/cMessageDrawer.h
+++ b/drawers/cMessageDrawer.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <string>
+
 class BITMAP;
 
 namespace messages {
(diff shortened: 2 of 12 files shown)


drawers/cMessageDrawer.h
```diff
@@ -32,15 +34,15 @@ class cMessageDrawer {
...
+		void setMessage(const std::string &msg);  // set a new message
+		const std::string& getMessage() { return cMessage; }
+		bool hasMessage() { return (!cMessage.empty()); }
 
 	protected:
 
 	private:
 		// MESSAGING / HELPER
-		char cMessage[255];
+		std::string cMessage;
```
pbomta, line 45: 👍

cGame_logic.cpp
```diff
@@ -1623,19 +1623,10 @@ void cGame::drawCombatMouse() {
...
-    } else if (m_screenshot < 100) {
-        sprintf(filename, "%dx%d_00%d.bmp", m_screenX, m_screenY, m_screenshot);
-    } else if (m_screenshot < 1000) {
-        sprintf(filename, "%dx%d_0%d.bmp", m_screenX, m_screenY, m_screenshot);
-    } else {
-        sprintf(filename, "%dx%d_%d.bmp", m_screenX, m_screenY, m_screenshot);
-    }
 
-    save_bmp(filename, bmp_screen, general_palette);
+    std::string filename = fmt::format("{}x{}_{:0>4}.bmp", m_screenX, m_screenY, m_screenshot);
```
pbomta, line 1627: 👍

gamestates/cSelectYourNextConquestState.cpp
```diff
@@ -111,10 +111,8 @@ void cSelectYourNextConquestState::thinkFast() {
         }
 
         cMessageDrawer *pDrawer = drawManager->getMessageDrawer();
-        char *cMessage = pDrawer->getMessage();
-
         isFinishedConqueringRegions = true;
-        bool isNotDisplayingMessage = cMessage[0] == '\0';
+        bool isNotDisplayingMessage = pDrawer->hasMessage();
```
pbomta, line 115: Logic inverterted here accidentally.
```suggestion
        bool isNotDisplayingMessage = !pDrawer->hasMessage();
```
or perhaps
```suggestion
        bool isDisplayingMessage = pDrawer->hasMessage();
```
And then invert the logic that uses the boolean.

utils/cLog.cpp
```diff
@@ -165,12 +165,12 @@ void cLogger::log(eLogLevel level, eLogComponent component, const std::string& e
...
     logline += event;
 
-    fprintf(m_file, "%s\n", logline.c_str()); // print the text into the file
-    fflush(m_file);
+    logCommentLine(logline);
 }
 
 void cLogger::logCommentLine(const std::string& txt) {
-    fprintf(m_file, "\\\\%s\n", txt.c_str()); // print the text into the file
+    m_file << txt << std::endl;
```
pbomta, line 172:
```suggestion
    m_file << "\\" << txt << std::endl;
```
This is some feature that Stefan likes. There are lines like this in the logfile:
```
\\-------
\\Allegro
\\-------
```

utils/cLog.cpp
```diff
@@ -165,12 +165,12 @@ void cLogger::log(eLogLevel level, eLogComponent component, const std::string& e
 
     logline += event;
 
-    fprintf(m_file, "%s\n", logline.c_str()); // print the text into the file
-    fflush(m_file);
+    logCommentLine(logline);
```
pbomta, line 168: The comment line function is to log text with two backslaches in front of it.
TBH, I'm not a big fan of it, but Stefan likes it to have "headers" before certain blocks of log lines.
What do you say, @stefanhendriks, remove this feature? We log the component with each log line, that may be enough.
I'm OK either way.

utils/cLog.h
```diff
@@ -65,7 +68,8 @@ class cLogger {
       cLogger(const cLogger&) = delete;
       cLogger(cLogger&&) = delete;
 
-      std::FILE* m_file;
+      //std::FILE* m_file;
```
pbomta, line 71: Let's just remove the commented out lines.

pbomta (changes requested):
Nice improvement!
If the logic is fixed and the comment line reinstated (or removed, if Stefan is OK with that), then it can go in.

utils/cLog.h
Akatsuki57: Done

utils/cLog.cpp
Akatsuki57: I fix as previously

utils/cLog.cpp
Akatsuki57: Oups. i fix as previously

pbomta (approved):

Akatsuki57:
This branch compile but she is unstable. I will check what's wrong !

utils/cLog.cpp
stefanhendriks: haha true. I like to have some 'headers' to tell when certain parts of logging start... so yeah please keep that ;-)

stefanhendriks:
nice work @Akatsuki57 ! Let us know when it is stable. I'll test it then on my machine and when it works I'll approve / merge.

stefanhendriks:
I made a draft of this PR so we know the status. If it is ready to go we can put it back with "Ready for review"

PR #464, opened 2022-02-19
proper includes for mentat and controls directories

pbomta:
After this PR and the other I opened with proper includes, only a few files remain that still include the big d2tmh.h header file. So close to its deletion... 😄


stefanhendriks:
nice! and good to see how we're getting closer and closer getting rid of that `d2tmh.h` file :) - nice going @pbomta !

Issue #473, opened 2022-02-20
Clicking on unit or structure should give information about it in top bar


Akatsuki57:
For example, i click on my siege tank and in this place rounded in red, we see a text information like :

Siege tank: 85% heal - on standby

![800x600_0000](https://user-images.githubusercontent.com/99535165/154861420-25946f2f-3e77-4b27-b695-f595718be505.png)

stefanhendriks:
oh that should be relatively easy to do actually. I'm working on the pieces of code as we speak, so let me see if I can make this work.

PR #477, opened 2022-02-20
Render message when unit is selected

stefanhendriks:
depends on #478 

- ie, one 1 unit on box-select
- 1 unit specifically selected
- structure selected

for #473


stefanhendriks:
Example: 
image

PR #482, opened 2022-02-21
refactor: remove the d2tmh master header

pbomta:
Also pragma once everywhere.
And forward declares BITMAP as a struct, as clang can choke on this.


stefanhendriks:
oh wow you went full force on this. I will look into this later. Nice achievement though!

stefanhendriks:
test it a bit, works great!

Issue #466, opened 2022-02-20
Proper order of bootstrapping d2tm so that we can read settings and startup flags with sane order


stefanhendriks:
Currently, the way d2tm is bootstrapped is all messed up. 

It means that resource allocation != instantiating (which we do want). But it also means we set values directly into a global `game` class from startup params, but then later overwrite from ini files.

This means, that config/ini files supersede startup parameters.

However, I think startup parameters should superside ini files instead.

In abstract sense I think this should become the picture:

- read startup params
- read config file
- merge two together to determine 'configuration state'
- initialize game with this 'configuration state' (ie, create unique ptr for game at this point with config object as constructor input)
- proceed...

Akatsuki57:
To do this, I'm going to build a third party class that will read command line options and store them in an internal map.
This map will contain the only options that we have analyzed.

Next, the program will read the ini files. After, I switch back to the options stored in the map.
The program continues as normal.

stefanhendriks:
The PR make sure the program arguments override the config in `game.ini` which is expected behavior.

PR #488, opened 2022-02-23
Feature/order of  bootstrapping

Akatsuki57:
Yet, command line options take precedence over ini file options.

New options : screenWidth, screenHeight

New param in game.ini : Fullscreen

Improve wrong param : program stop


stefanhendriks:
Thanks for this PR!

I could verify the params working, overriding the `game.ini` setting. Which is great.

I have added `--help` switch to print out available instructions.
Also when unknown argument is passed, it will print out that argument.

A little feedback: when naming variables, try to have them start with `m_`, we're trying to get this in consistently. I've corrected this myself in this PR; just for future reference.

Other than that looks good to me. 👍 

I think this is good for now. 

Will merge when green.

Example when executing `d2tm.exe --help`:
image

stefanhendriks:
This PR is for #466

Issue #480, opened 2022-02-20
Make AI a bit more aggressive


stefanhendriks:
When playtesting skirmish I found the skirmish AI very slow/passive.

I think because of two reasons:

- [x] the AI does not build up a proper economy at start, hence running out of funds too soon
- [ ] the AI is not able to gauge how their economy is properly, hence staying too long in an invalid economy state and thus not building forces properly
- [ ] ... ?

PR #491, opened 2022-02-27
Make AI less passive

stefanhendriks:
for #480


stefanhendriks:
I noticed that the AI had a few bugs; fixed them. That helped a lot already. Made the way how AI deals with economy a bit different. Also made it a bit more random how/when to build additional units. 

Seems to be a lot more aggressive/fun now. Although it will require more playtesting. These things take considerable amount of time to test.

PR #495, opened 2022-03-03
Chore/rewriting map

Akatsuki57:
- Rewrite loading preview map without allegro
- Fix memory leak


stefanhendriks:
build fails, because of reasons. Need to check that, after it is green again this can be merged.

AssociationSirius:
I didn't have any error when i have work on this PR.
I can't see because it's on Windows-msys2

mnhauke:
> I didn't have any error when i have work on this PR. I can't work on it because it's on Windows-msys2

I can see the same compilation error on Linux with gcc 12.1. It seems related to some changes in gcc 12.x
https://gcc.gnu.org/gcc-12/changes.html
> [-Waddress](https://gcc.gnu.org/onlinedocs/gcc-12.1.0/gcc/Warning-Options.html#index-Waddress) has been enhanced so that it now warns about, for instance, comparing the address of a nonstatic member function to null ([PR102103](https://gcc.gnu.org/PR102103))

AssociationSirius:
My bad, my code, i will find a solution ...

AssociationSirius:
> > I didn't have any error when i have work on this PR. I can't work on it because it's on Windows-msys2
> 
> I can see the same compilation error on Linux with gcc 12.1. It seems related to some changes in gcc 12.x https://gcc.gnu.org/gcc-12/changes.html
> 
> > [-Waddress](https://gcc.gnu.org/onlinedocs/gcc-12.1.0/gcc/Warning-Options.html#index-Waddress) has been enhanced so that it now warns about, for instance, comparing the address of a nonstatic member function to null ([PR102103](https://gcc.gnu.org/PR102103))

Thx !

stefanhendriks:
I could reproduce it with gcc12 as well (msys2 on windows); made a small change and pushed it - it should compile now.

PR #496, opened 2022-03-04
Rewriting/c preview maps

Akatsuki57:
After merge #490 and #495 :
- class creation cPreviewMaps that includes s_PreviewMap and all the functions associated with it


stefanhendriks:
Found the reason. 

It fails because a filename (`4PL_SinéadPlains.ini`) is not being found. This is due to https://stackoverflow.com/questions/5727791/opening-a-text-file-with-fstream-but-filename-characters-are-not-in-ascii/5727958 

(ie , `std::ifstream` cannot deal with unicode chars)

stefanhendriks:
Found another bug in this PR. The `startcell` items in an `ini` file are not with comma separated values (ie, it is not `StartCell=1234,456,432`). Instead, the ini file contains multiple `StartCell=1234` lines.

I'll change the code so that works again.

PR #505, opened 2022-03-21
Rewriting/fmt  format

Akatsuki57:
Eradication of printf in several parts of the program with the power of fmt::format


stefanhendriks:
I squashed all 4 commits into one and rebased on master

PR #573, opened 2022-10-23
FMT version 9.0.0 integration

Mira-Chronos:
I transformed `char[64]`, `char[256]` into `string` but the program crashed. the debugger return me to fmt::format.
I investigated: it is about enum class which does not pass under fmt. I took the opportunity to upgrade fmt to version 9. This version give me clear and definite errors. I fixed them. Yet I can continue my work on externalizing game data to another branch.


Mira-Chronos:
With C++20 with std::format

PR #575, opened 2022-10-25
Improvement data loading

Mira-Chronos:
As for the bullets, I outsourced the reading of data from SANDWORM and ORNITHOPTER. This work is temporary, it is a test on two units which, if successful, should be generalized to the other units.
I am waiting for the validation of this prototype before continuing the migration.

I am waiting for the validation of this prototype before continuing the migration.
There are people to deal with :)

- The creation of units.ini is temporary, it allows me to see what needs to be converted.
- I will make a special dataLoader class to relieve the common.cpp file
- The idea would be to read the datas with a for loop, which will be possible when I can easily access the names of the textures


Mira-Chronos:
To long opened, unusable yet.

Mira-Chronos:
Using with SDL2

PR #598, opened 2025-06-15
All sources files in src

Mira-Chronos:
I was going crazy with source files scattered everywhere.

I moved all the source files to src/, leaving the other files in the same locations.

I patched CMake to adapt to the new configuration.

I haven't supervised any scripts on Windows, so I need help on the Windows platform.


Mira-Chronos:
Done i SDL2 branch

PR #630, opened 2025-07-16
Add Mira to credits

Mira-Chronos:


src/gamestates/cCreditsState.cpp
```diff
@@ -265,6 +271,12 @@ void cCreditsState::prepareCrawlerLines()
         .color = Color::white(),
         .height = fontHeightWithALittlePadding
     });
+    m_lines.push_back(s_CreditLine {
+        .name = "Mira",
+        .txt = "for port of D2TM to SDL2",
```
stefanhendriks:
```suggestion
        .txt = "for porting D2TM to SDL2",
```

src/gamestates/cCreditsState.cpp
```diff
@@ -307,6 +319,12 @@ void cCreditsState::prepareCrawlerLines()
         .color = Color::white(),
         .height = fontHeightWithALittlePadding
     });
+    m_lines.push_back(s_CreditLine {
+        .name = "Mira",
+        .txt = "for the pride of working on a big project in C++",
```
stefanhendriks: Perhaps we should add a separate section for special messages by contributors; now your name would appear twice under "special thanks". I think this deserves a special place; you can leave a personal note to the audience then? wdyt?

src/gamestates/cCreditsState.cpp
Mira-Chronos: Sorry for my poor english

src/gamestates/cCreditsState.cpp
Mira-Chronos: Delete my last comment; it should not appear in contributions.
Contributing means contributing to a project. This is not some kind of thank you.

src/gamestates/cCreditsState.cpp
stefanhendriks: no probs! this is what PR reviews are for ;-)

src/gamestates/cCreditsState.cpp
stefanhendriks: Ok, I'll make something out of it. Perhaps we should have some special section with "messages for the audience by our contributors" ?

Mira-Chronos: Yes, this section will be useful !

Mira-Chronos:
@stefanhendriks new version, ready for merging !

PR #631, opened 2025-07-16
Cmake output

Mira-Chronos:
PR:
- CMake handle the version of D2TM
- Remove cGame.m_version
- No need to grab code to find version


CMakeLists.txt
```diff
@@ -5,6 +5,14 @@ set(CMAKE_CXX_STANDARD 20)
 project(d2tm)
 
 #SET(OpenGL_GL_PREFERENCE "GLVND")
+SET(D2TM_VERSION "0.7.0")
```
stefanhendriks: can we get this from an `environment variable` ? And if it does not exist fall back to a default value? That would make it possible to start producing binaries from GH Actions with tags for example.

stefanhendriks (commented):
Awesome, just need to fix the github action for linux (we can probably set the desired GCC version in the image) and add ability to set version via env vars.

CMakeLists.txt
Mira-Chronos: I don't know. I'm just showing you this possibility that saves you from digging through the code, and at the same time an elegant way to communicate from CMake to the program.

Mira-Chronos: Good new: after researches, we can use `environment variable`
CMake allow this: `$ENV{VARIABLE_NAME}`
You can write this `SET(D2TM_VERSION "$ENV{D2TM_VERSION_ENV}")`

stefanhendriks: Awesome, I hope it has a default fallback, ie, I'm used to notation like this `${D2TM_VERSION_ENV:<some default value>}`. If there is something equivalent it would be nice.

stefanhendriks:
> Linux compilation shows an error (even though I developed on Linux without error). It seems that g++ doesn't have the correct version on Github because this practice is allowed in C++20.

I have rebased this upon the latest branch which set version to `gcc14`, it works on my machine, will look into this further if it keeps failing.

PR #632, opened 2025-07-16
Refactor: Change the way how settings.ini is read and applied within the game

Mira-Chronos:
- There is a structure that fully manages all the game's options.
- This structure is first completed by reading an ini file.
- It is then modified by any command-line directives issued.
- Finally, it is applied to cGame.

### DIFF
--- a/src/cGame_logic.cpp
+++ b/src/cGame_logic.cpp
@@ -51,7 +51,6 @@
 #include "utils/Color.hpp"
 
 #include "utils/cFileValidator.h"
-#include "utils/cHandleArgument.h"
 #include "utils/cIniFile.h"
 #include "player/cHousesInfo.h"
 #include "utils/Graphics.hpp"
@@ -87,10 +86,9 @@ cGame::cGame() : m_timeManager(*this)
     m_version = "0.7.0";
 
     m_mentat = nullptr;
-    // m_handleArgument = std::make_unique<cHandleArgument>(this);
 }
 
-void cGame::applySettings(const GameSettings& gs)
+void cGame::applySettings(const GameSettings &gs)
 {
     m_screenW = gs.screenW;
     m_screenH = gs.screenH;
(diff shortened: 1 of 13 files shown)


src/cGame_logic.cpp
```diff
@@ -87,9 +86,30 @@ cGame::cGame() : m_timeManager(*this)
     m_version = "0.7.0";
 
     m_mentat = nullptr;
-    m_handleArgument = std::make_unique<cHandleArgument>(this);
 }
 
+void cGame::applySettings(const GameSettings &gs)
```
stefanhendriks: question: do we want to use a reference here? Or perhaps `std::shared_ptr` like before?

I am not entirely confident about pointer usage (or std's equivalent), so thats why I am asking.

src/cGame_logic.cpp
```diff
@@ -947,6 +934,7 @@ bool cGame::setupGame()
     }
     else {
         logger->log(LOG_INFO, COMP_INIT, "Load data", "Hooked datafile: " + settingsValidator->getName(eGameDirFileName::GFXDATA), OUTC_SUCCESS);
+        // memcpy(general_palette, gfxdata[PALETTE_D2TM], sizeof general_palette);
```
stefanhendriks, line 940: can be removed?

src/main.cpp
```diff
@@ -65,18 +68,22 @@ TTF_Font *bene_font;	// benegesserit font.
...
-    if (game.handleArguments(argc, argv) < 0) {
-        return 0;
-    }
+    // read file config
+    std::shared_ptr<cIniFile> settingsIniFile = std::make_shared<cIniFile>("settings.ini", true);
+    GameSettings settings = loadSettingsFromIni(settingsIniFile.get());
+    //read prompt config
+    cHandleArgument handleArg;
+    int parseResult = handleArg.handleArguments(argc, argv, settings);
+    if (parseResult < 0) return 1; // Affiche l'aide ou erreur
```
stefanhendriks: `Affiche l'aide ou erreur` ? Perhaps print out some message that we can't parse `settings.ini`.

src/utils/common.cpp
```diff
@@ -22,9 +22,37 @@
 #include "utils/cSoundPlayer.h"
 #include "utils/cIniFile.h"
 #include "utils/Graphics.hpp"
+#include "utils/GameSettings.hpp"
 
 #include <cmath>
 
+GameSettings loadSettingsFromIni(cIniFile *settings)
```
stefanhendriks: here we use a raw pointer, is that wise to do? perhaps use that shared pointer we used elsewhere?

src/main.cpp
Mira-Chronos: Oki, it will be done. Thx for reply.

src/cGame_logic.cpp
Mira-Chronos: okay, i will modify it.

src/utils/common.cpp
Mira-Chronos: yes, i use a raw pointer. The reason is simple: if I use a raw pointer, it's because the management of its initialization/destruction is left to another entity.

As in other parts of my code, which I plan to generalize, I'd like to convert all raw pointers that manage the object to smart pointers and leave all raw pointers that use the object as raw. (Otherwise, we'd terraform the source code.)

I see a unique ptr as the responsibility of the class that created it. Shared_pointer is already a group of classes that manage a resource. A group uses it, not a specific entity.
Weak ptr is used to avoid loops in declarations, for example.

The software as it stands doesn't crash; the introduction of smart pointers would allow for greater automatic resource release.

stefanhendriks: so basically, the owner (the one who created the pointer) will have a `shared_pointer`, and the non-owners have a raw pointer, do I understand correctly?

I'd have to read up on shared/unique/raw pointers to get a better understanding; all I do know is that having a std wrapper makes it easier to do your resource cleanup. (taken care for you).

Mira-Chronos: Yes ! 
Owner use unique_ptr (if it use alone.)
Owners will share shared_ptr (if used by many)
Weak_ptr is for break cycles with shared_ptr
raw pointer : no possession

A smart pointer manages its memory automatically.
There are no more memory leaks.

Documentation here you have : 
https://en.cppreference.com/book/intro/smart_pointers

src/cGame_logic.cpp
Mira-Chronos: it's removed

PR #633, opened 2025-07-17
Let rnd() implementation use std::uniform_int_distribution

Mira-Chronos:
This PR replaces the current `rnd()` function with a new RNG::rnd() using https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution.html

This is a more modern way of generating random numbers.

The uniform implementation has less bias, [see a possible explanation here](https://stackoverflow.com/a/32860886).

### DIFF
--- a/src/utils/RNG.hpp
+++ b/src/utils/RNG.hpp
@@ -3,28 +3,13 @@
 
 class RNG {
 public:
-    static std::mt19937& getGenerator() {
-        static std::random_device rd;
-        static std::mt19937 s_generator(rd());
-        return s_generator;
-    }
+    static std::mt19937& getGenerator();
 
-    static int genInt(int min, int max) {
-        std::uniform_int_distribution<int> dist(min, max);
-        return dist(getGenerator());
-    }
+    static int genInt(int min, int max);
 
-    static int rnd(int max) {
-        if (max<1)
-            return 0;
-        std::uniform_int_distribution<int> dist(0, max);
-        return dist(getGenerator());
-    }
+    static int rnd(int max);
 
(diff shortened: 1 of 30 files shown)


src/utils/RNG.hpp
```diff
@@ -14,6 +14,13 @@ class RNG {
         return dist(getGenerator());
     }
 
+    static int rnd(int max) {
+        if (max<0)
```
stefanhendriks: this is slightly different, as long as the new code can deal with a range between `0-0` and won't crash this should be 👌

src/utils/d2tm_math.cpp
```diff
@@ -220,11 +220,11 @@ int convertAngleToDrawIndex(int faceAngle, bool clockWiseBitmap, int offset, int
 }
 
 // return random number between 0 and 'max'
-int rnd(int max)
-{
-    if (max < 1) return 0;
-    return std::rand() % max;
-}
+// int rnd(int max)
```
stefanhendriks: if it is no longer used, we can delete this code

src/utils/d2tm_math.h
```diff
@@ -10,7 +10,7 @@ int convertAngleToDrawIndex(int faceAngle, bool clockWiseBitmap = false, int off
 // bullet only
 int bullet_face_angle(float angle);
 
-int rnd(int max);
+// int rnd(int max);
```
stefanhendriks: same here

stefanhendriks:
Good that you merged it, last comment, try to `rebase` master into branches. This gives a cleaner git history.

PR #638, opened 2025-07-22
New GuiButton

Mira-Chronos:
Simplified use of GuiButtons and applications

- GuiButton has a Builder
- GuiButton can display an image instead of text
- Centralized GuiButton colors with themes
- Created Labels and Checkboxes adapted from GuiButton
- Possible use of lambda functions
- Implementation of smarts_pointers
- Mentats classes now have real GuiButtons


src/cGame_logic.cpp
```diff
@@ -2180,4 +2180,65 @@ void cGame::onKeyDownDebugMode(const cKeyboardEvent &event)
 void cGame::setMousePosition(int w, int h)
 {
     m_mouse->setCursorPosition(window, w,h);
-}
\ No newline at end of file
+}
+
+void cGame::execute(AbstractMentat &mentat)
+{
+    if (game.isState(GAME_BRIEFING)) {
```
stefanhendriks, line 2195: it would be even more cool if this was somewhere provided when constructing the mentat class. Ie, in case of campaign mission inject behavior to get to the next mission. Else, for skirmish do other things. Wdyt? (what do you think)

src/controls/cGameControlsContext.h
```diff
@@ -62,6 +62,9 @@ class cGameControlsContext : public cInputObserver, cScenarioObserver {
     bool shouldDrawToolTip() const {
         return m_drawToolTip;
     }
+    bool isMouseRightButtonPressed() const{
```
stefanhendriks, line 65: preferably this goes via `event`s , why need this?

src/gamestates/cMainMenuGameState.cpp
```diff
@@ -50,10 +41,17 @@ cMainMenuGameState::cMainMenuGameState(cGame &theGame) : cGameState(theGame), te
 
     int creditsX = (game.m_screenW / 2) - buttonWidth;
     const cRectangle &creditsRect = cRectangle(creditsX, 0, buttonWidth, buttonHeight);
-    gui_btn_credits = new cGuiButton(textDrawer, creditsRect, "CREDITS", buttonKinds);
-    gui_btn_credits->setTextAlignHorizontal(buttonTextAlignment);
-    gui_btn_credits->setOnLeftMouseButtonClickedAction(new cGuiActionToGameState(GAME_CREDITS, true));
 
+    gui_btn_credits = GuiButtonBuilder()
```
stefanhendriks, line 46: I *love* the builder pattern. Nice to see it here!

src/gamestates/cMainMenuGameState.cpp
```diff
@@ -50,10 +41,17 @@ cMainMenuGameState::cMainMenuGameState(cGame &theGame) : cGameState(theGame), te
...
+            .withRect(creditsRect)        
+            .withLabel("CREDITS")
+            .withTextDrawer(&textDrawer)    
+            .withTheme(GuiTheme::Light())
+            .withKind(GuiRenderKind::TRANSPARENT_WITHOUT_BORDER)
+            .onClick([this]() {
```
stefanhendriks, line 52: oh this is nice, I've never seen lambda's in C++ before like this. Seems more compact than last time I saw it.

src/gamestates/cMainMenuGameState.cpp
```diff
@@ -63,81 +61,113 @@ cMainMenuGameState::cMainMenuGameState(cGame &theGame) : cGameState(theGame), te
...
+            .build();
+    gui_window->addGuiObject(std::move(gui_btn_SelectHouse));
 
     int skirmishY = 344;
     const cRectangle &skirmish = cRectangle(buttonsX, skirmishY, buttonWidth, buttonHeight);
-    cGuiButton *gui_btn_Skirmish = new cGuiButton(textDrawer, skirmish, "Skirmish", buttonKinds);
-    gui_btn_Skirmish->setTextAlignHorizontal(buttonTextAlignment);
-    gui_btn_Skirmish->setOnLeftMouseButtonClickedAction(new cGuiActionSetupSkirmishGame());
-    gui_window->addGuiObject(gui_btn_Skirmish);
+    std::unique_ptr<GuiButton> gui_btn_Skirmish = GuiButtonBuilder()
```
stefanhendriks, line 89: not sure if this is possible, but in my other projects I am used to this kind of reference to the Builder:

`ThingToBuild.builder().withX().withY().build()`. 

Could you change it so that this line reads like:
```suggestion
    std::unique_ptr<GuiButton> gui_btn_Skirmish = GuiButton.builder()
```
?

src/gamestates/cOptionsState.cpp
```diff
@@ -127,7 +141,8 @@ void cOptionsState::setPrevState(int prevState)
 
 void cOptionsState::refresh()
 {
-    delete m_guiWindow;
-    m_guiWindow = nullptr;
+    // delete m_guiWindow;
```
stefanhendriks: commented out code can be removed

src/cGame_logic.cpp
```diff
@@ -1338,17 +1338,17 @@ void cGame::createAndPrepareMentatForHumanPlayer(bool allowMissionSelect)
     delete m_mentat;
     int houseIndex = players[HUMAN].getHouse();
     if (houseIndex == ATREIDES) {
-        m_mentat = new cAtreidesMentat(allowMissionSelect);
+        m_mentat = new AtreidesMentat(allowMissionSelect);
```
stefanhendriks, line 1339: for consistency sake I'd keep the `c` Prefix , so `cAtreidesMentat` is fine.

If you really want to take this a step further, we could argue the mentat being constructed can be a composited version. (ie, `cMentat(<with some properties>`) so it reflects an AtreidesMentat, that wat the `atreides` part would become a factory method, and no longer a concrete implementation).

src/cGame_logic.cpp
Mira-Chronos: I remember putting this bunch of code here because the lambda function was getting complex to write and I didn't want to rewrite everything, because with the future state branch, this code would no longer have its place here and would be completely rewritten. Let's say it's temporary. 
I'm not very familiar with how the software knows where the player is in the missions, but it's true that an eventHandler would be more judicious.

src/cGame_logic.cpp
Mira-Chronos: As spoken on Discord

src/controls/cGameControlsContext.h
Mira-Chronos: I create it for future to implement drawToolTip that is broken because i didn't have key access yet on this part of code. I wasn't sure how I could then link it to events.

src/gamestates/cMainMenuGameState.cpp
Mira-Chronos: It was a goal of this branch !

src/gamestates/cMainMenuGameState.cpp
Mira-Chronos: Lambda functions have evolved a lot since C++11

PR #642, opened 2025-07-23
Fix shadow rendering issues after SDL2 migration

Mira-Chronos:
Restores shadow rendering on in-game units and buildings


src/include/structs.h
```diff
@@ -100,6 +101,7 @@ struct s_FlagInfo {
 struct s_StructureInfo {
     // Structures
     SDL_Surface *bmp;         // a pointer to the 8 bit bitmap
+    Texture *shadow;	   // a pointer to the shadow bitmap
```
stefanhendriks, line 104: Trying to understand this, why does this need to be a texture? And not an `SDL_Surface` ?

src/gameobjects/units/cUnit.cpp
```diff
@@ -735,6 +735,36 @@ void cUnit::draw()
...
+        SDL_FreeSurface(shadow);
+        SDL_FreeSurface(stretchedShadow);
+    }*/
+    Texture *shadow = cPlayer.getUnitShadowBitmap(iType);
+    if (shadow) {
+        cRectangle src = {start_x, start_y, bmp_width, bmp_height};
+        if (iType == CARRYALL) {
+            // start_x += 2;
+            // start_y += 2;
+            src = {start_x+2, start_y+2, bmp_width, bmp_height};
```
stefanhendriks: so the translation between unit/body facing and which shadow 'piece' to draw happens here now, (instead in `getUnitShadowBitmap`). For now we can keep it there.

If this works btw, we can get rid of the commented out code above.

src/player/cPlayer.cpp
```diff
@@ -256,16 +256,16 @@ void cPlayer::setHouse(int iHouse)
         emblemBackgroundColor = getEmblemBackgroundColorForHouse(house);
 
         destroyAllegroBitmaps();
-        bmp_flag = createTextureFromIndexedSurfaceWithPalette(gfxdata->getSurface(BUILDING_FLAG_LARGE),232);
-        bmp_flag_small = createTextureFromIndexedSurfaceWithPalette(gfxdata->getSurface(BUILDING_FLAG_LARGE),232);
+        bmp_flag = createTextureFromIndexedSurfaceWithPalette(gfxdata->getSurface(BUILDING_FLAG_LARGE),IndTrans);
```
stefanhendriks: Not sure what `IndTrans` means, perhaps we can rename it to something more descriptive? I didn't know there was so much transparency happening to be honest. 😅 . 

Suggestion:

`DefaultTransparency`

src/player/cPlayer.cpp
```diff
@@ -520,6 +520,42 @@ Texture *cPlayer::getUnitTopBitmap(int index)
...
+// SDL_Surface *shadow = SDL_CreateRGBSurface(0, bmp_width, bmp_height,32,0,0,0,255);
+// renderDrawer->FillWithColor(shadow, Color{255,0,255,255});
+// renderDrawer->blit(sUnitInfo[index].shadow, shadow, start_x, start_y, 0, 0, bmp_width, bmp_height);
+// return shadow;
+// }
+// return nullptr;
+// }
+Texture *cPlayer::getUnitShadowBitmap(int index)
+{
+    if (sUnitInfo[index].shadow) {
```
stefanhendriks, line 552: any reason why the translation got moved out of there? Curious.

src/include/structs.h
Mira-Chronos: `SDL_Surface` are used from CPU, when using you must transfer it from CPU to GPU, every time. `Texture` already are in GPU so no link, no transfer. Really smart and easy to use.

src/gameobjects/units/cUnit.cpp
Mira-Chronos: I keep commented code to explain where this code comes from.

**long story:**
Previously, we would cut the sprites in the CPU memory and then send the piece to the display.
Currently, the entire image of all the sprites is in the GPU, and we ask the GPU to cut the sprite we're interested in.
This is faster and more convenient. Therefore, the old method returns the entire sprite, not a piece.

src/player/cPlayer.cpp
Mira-Chronos: Its exact name is

`Transparent color index of the 8-bit RGB palette`

I'm poor to give his definitive real name  ! Help me !

src/player/cPlayer.cpp
Mira-Chronos: It is linked to the long history described above.

stefanhendriks:
The 'move order' indication should also be transparent (see 0.7.0), this is still opaque.

Issue #659, opened 2025-08-01
global_map on cMap


Mira-Chronos:
oh let see with in cMap:

```
bool cMap::occupiedByWallOrMountain(int iCell)
{
    if (this != &global_map) {
        std::cerr << "using occupiedByWallOrMountain with other instance on global_map !" << std::endl;
    }
    if (iCell < 0 || iCell >= maxCells) return false;

    if (global_map.getCellType(iCell) == TERRAIN_WALL) return true;
    if (global_map.getCellType(iCell) == TERRAIN_MOUNTAIN) return true;

    return false;
}
```

test class without global
remove global_map if safe

Mira-Chronos:
<img width="472" height="131" alt="Image" src="https://github.com/user-attachments/assets/e984577b-3b83-418d-9555-dc6df24efcf2" />

Not only once

PR #661, opened 2025-08-01
3_ Iteration on cMaps

Mira-Chronos:
The goal was to create MapGeometry to relieve cMap of all geometry calculations.
This allows to use only a targeted part of cMap's functionality, in places where only geometry was needed.

- Create MapGeometry
- Remove `global_map` from `cMap` after multiple tests.
- Discorver #659 !!!

This branch allows you to truly break down cMap and make it more testable.

However, I'm missing the `ctx` branch to generalize its deployment. Often, I only need MapGeometry and not the entire global_map, which can be easily resolved with `ctx` and is currently impossible for me.

### DIFF
--- a/src/map/cMap.cpp
+++ b/src/map/cMap.cpp
@@ -54,9 +54,9 @@ cMap::~cMap()
     }
 }
 
-MapGeometry *cMap::getGeometry() const
+MapGeometry &cMap::getGeometry() const
 {
-    return mapGeometry.get();
+    return *mapGeometry;
 }
 
 void cMap::setReinforcements(std::shared_ptr<cReinforcements> reinforcements)
@@ -1036,7 +1036,7 @@ int cMap::findNearestSpiceBloom(int iCell)
 
     if (iCell < 0) {
         // use cell at center
-        iCell = getGeometry()->getCellWithMapDimensions(halfWidth, halfHeight);
+        iCell = getGeometry().getCellWithMapDimensions(halfWidth, halfHeight);
         iDistance = getWidth();
     }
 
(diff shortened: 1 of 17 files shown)


src/map/cMap.cpp
```diff
@@ -52,6 +54,11 @@ cMap::~cMap()
     }
 }
 
+MapGeometry *cMap::getGeometry() const
+{
+    return mapGeometry.get();
```
stefanhendriks: I think we should change this to return the reference, and not a pointer. As it will *always* be available from the map. This makes the code from the caller perspective easier to use.

Ie:

```cpp
MapGeometry &cMap::getGeometry() const
{
    return *mapGeometry;
}
```

src/map/cMap.cpp
```diff
@@ -98,6 +105,7 @@ void cMap::init(int width, int height)
 
     this->width = width;
     this->height = height;
+    mapGeometry->resize(this->width,this->height);
```
stefanhendriks, line 108: Thinking about this; this is a more of a set of handy calculations based on width/height of a 2d plane. We could also re-create the `mapGeometry` class, if we want to stick to more Functional Programming principles. The upside is that we don't need to worry about changing state. The downside is the (re)creation of the MapGeometry class.

src/map/cMap.cpp
```diff
@@ -188,12 +196,12 @@ bool cMap::canDeployUnitTypeAtCell(int iCell, int iUnitType)
 
     if (isWorm) {
         int wormId = getCellIdWormsLayer(iCell);
-        return global_map.isCellPassableForWorm(iCell) && wormId < 0;
+        return isCellPassableForWorm(iCell) && wormId < 0;
```
stefanhendriks, line 199: Good find 👍  - no need to rely on `global_map` if we can access internal information.

src/map/cMap.cpp
stefanhendriks: I changed this myself.

src/map/cMap.cpp
stefanhendriks: for now this is fine.

PR #662, opened 2025-08-02
4 Iteration on Map Generator

Mira-Chronos:
Add the ability to create random maps with different resolutions 
- unify random map creation i one area in code
- customizable random map size
- remove old progress bar, unusable with SDL2


stefanhendriks:
I noticed that the rocket turrets and the orni thopter sounds are swapped. I played as Ordos.

stefanhendriks:
> Start positions got broken because of the map format changes i think Mira was doing. He asked me to correct the starting cells manually but with the current editor, loading maps and making changes isn't really possible without extreme hacky workarounds
> 
> its typically with non-square maps where the coordinates are offset by some amount.

Hmm, they seem to be off by quite a lot. Afaik the calculation/map reading had some impact (because of a border, or not, now taken into account. So I did not expect this to be off that much. I'll look into it.

Regardless, cell calculation itself works fine, the game runs and all other coordination specific things seem to work. And the loading of the preview maps seem to directly take the value of the file, so there does not seem to be something wrong there. That seems to indicate, as you say, that the file does not contain the proper cell value.

Ripps1:
I think at a later date i started adding buffer tiles to the map on the right and bottom edges to allow me to work with a central tile but i couldn't tell you exactly when that happened or which maps it applies to

at some point I'd honestly like to remake each of the maps with this boarder in mind and at the same time make improvements and corrections I couldn't before due to problems with the editor being unable to load maps correctly or at all.

stefanhendriks:
So if I am being naive, we can simply get rid of the last cells (trim the last character) and remove the 63th line in the ini file. That *would* fix it, unless the 1st char is skipped (so trim the left char instead). I need to look into this a bit more.

In code we also seem to take 65x65 into account (as we add 2 on top of the amount of cells we detect (for width and height), and since we have 63x63 stored, we end up with 65x65 in the internal data structure)

stefanhendriks:
As discussed, the maps are fixed [in this PR](https://github.com/stefanhendriks/Dune-II---The-Maker/pull/678), so I'll merge this one.

PR #698, opened 2025-11-01
Fix for #692 - Particles have House color (corpses should render with proper house color)

### LINKED ISSUE #692: Corpes always red
After SDL2 port, Corpes of dying infantry are red. (Harkonnen stuff)
It should be the same color of his House

Mira-Chronos:
Fix to this:
close #692 

image

### DIFF
--- a/src/gameobjects/particles/cParticle.cpp
+++ b/src/gameobjects/particles/cParticle.cpp
@@ -825,6 +825,8 @@ void cParticle::recolorForHouseIfGiven() {
         bmp = recoloredBmp;
         bmpOwner = true;
     } else {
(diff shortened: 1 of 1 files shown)


stefanhendriks:
I refactored a tiny bit, and then ran into a segfault. Looks like squishing infantry causes an error because we provide a `iHouse`, but it is actually the `iPlayer` and is not something we can recolor. Making the code be able to deal with that and print sensible log messages. (I want recoloring of particles to be flexible so we can chose what bitmaps to recolor in the future)

stefanhendriks (approved):

PR #703, opened 2025-11-02
rewriting/Study of enum classes

Mira-Chronos:
Test of rewriting some define in game
- observation of the consequences
- discovery of https://github.com/Neargye/magic_enum

All commented lines aren't erased because they show me that to do with them

Some code improvements

### DIFF
--- a/src/utils/d2tm_math.cpp
+++ b/src/utils/d2tm_math.cpp
@@ -22,7 +22,6 @@
 #include "common.h"
 
 #include <cassert>
-#include <cmath>
 #include <cstdlib>
 
 // returns in fRadians
--- a/src/utils/d2tm_math.h
+++ b/src/utils/d2tm_math.h
@@ -1,5 +1,8 @@
 #pragma once
 
+// Taken from cmath.h - An easy way to be cross-platform compatible, see also https://stackoverflow.com/a/6563891/214597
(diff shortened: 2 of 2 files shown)


src/drawers/cMapDrawer.cpp
```diff
@@ -148,8 +148,8 @@ void cMapDrawer::drawTerrain()
             }
 
             // draw Smudge if necessary
-            if (cell->smudgetype > -1 && cell->smudgetile > -1) {
-                const cRectangle src_pos = {cell->smudgetile * 32, cell->smudgetype * 32,32, 32};
+            if (cell->smudgetype.has_value() && cell->smudgetile > -1) {
+                const cRectangle src_pos = {cell->smudgetile * 32, static_cast<int>(cell->smudgetype.value()) * 32,32, 32};
```
stefanhendriks: Since I work with `Optional` in Java code bases, I wonder if we can take this a step further. Will experiment.

stefanhendriks: found a way to do this in a more functional manner @Mira-Chronos :-)

stefanhendriks:
https://github.com/Neargye/magic_enum looks interesting, no need to write own `toString` (and possibly forget it). Did you want to include it in this PR?

PR #712, opened 2025-11-07
Rewriting text 2 / Remove local variables

Mira-Chronos:
Remove local variables

### DIFF
--- a/src/cGame_logic.cpp
+++ b/src/cGame_logic.cpp
@@ -856,7 +856,7 @@ bool cGame::setupGame()
     ctx->setGraphicsContext(context->createGraphicsContext());
     // share Text to all class what use ctx !
     ctx->setTextContext(context->createTextContext());
-    global_map.getContext(ctx.get());
+    global_map.setGameContext(ctx.get());
 
     m_textDrawer = ctx->getTextContext()->gameTextDrawer.get();
     m_textDrawer->setApplyShadow(false);
--- a/src/map/cMap.cpp
+++ b/src/map/cMap.cpp
@@ -61,7 +61,7 @@ MapGeometry &cMap::getGeometry() const
     return *mapGeometry;
 }
 
-void cMap::getContext(GameContext* ctx)
+void cMap::setGameContext(GameContext* ctx)
 {
     m_ctx = ctx;
     m_textDrawer = ctx->getTextContext()->beneTextDrawer.get();
--- a/src/map/cMap.h
+++ b/src/map/cMap.h
@@ -57,7 +57,7 @@ public:
     **/
     double distance(int cell1, int cell2);
 
-    void getContext(GameContext* ctx);
+    void setGameContext(GameContext* ctx);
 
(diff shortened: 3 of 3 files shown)


src/context/ContextCreator.cpp
```diff
@@ -71,6 +71,13 @@ ContextCreator::ContextCreator(SDL_Renderer *renderer, cFileValidator *settingsV
     }  
 }
 
+ContextCreator::~ContextCreator()
+{
+    TTF_CloseFont(game_font);
```
stefanhendriks, line 76: @Mira-Chronos I checked and I assume that TTF_CloseFont() (which calls SDL_Free()) will eventually do the same thing as `delete game_font` in a sense.

src/drawers/cBuildingListDrawer.cpp
```diff
@@ -12,7 +12,8 @@
 #include <format>
 
 cBuildingListDrawer::cBuildingListDrawer(const GameContext *ctx, cPlayer *player) :
-    m_textDrawer(game_font),
+    m_gameTextDrawer(ctx->getTextContext()->gameTextDrawer.get()),
```
stefanhendriks, line 15: nice @Mira-Chronos 👌

src/drawers/cMouseDrawer.cpp
```diff
@@ -109,25 +109,24 @@ void cMouseDrawer::drawToolTip()
...
 
     if (context->isMouseOverStructure()) {
-        auto textWriter = cTextWriter((x + 2), (y + 2), small_font, 12);
-
+        textWriter->moveTo(x + 2, y + 2);
         cAbstractStructure *theStructure = context->getStructurePointerWhereMouseHovers();
 
         drawToolTipBackground();
-        drawToolTipGeneralInformation(theStructure, textWriter);
+        drawToolTipGeneralInformation(theStructure);
```
stefanhendriks, line 116: interesting, we do not draw tooltip as the method for setting `m_drawToolTip` to `true` is uncommented. (checking key `T`).

src/context/ContextCreator.cpp
Mira-Chronos: I fixed this good place in text PR3

src/drawers/cMouseDrawer.cpp
Mira-Chronos: I'm having trouble putting the right components in that part of the code. I have it in mind and I'll fix it when ctx is around.

stefanhendriks: basically we need to fetch the T key event and set the variable somewhere; we should set it from the game loop somewhere into the mouseDrawer (from playable state only). But we'll fix that later, yes.

src/context/ContextCreator.cpp
stefanhendriks: I'll check again then, as when I run (and quit) the game, I end up in a SEGFAULT

image

PR #716, opened 2025-11-08
Rewriting text 3 / cTextDrawer has now a texture cache to avoid waste CPU resources

Mira-Chronos:
- cTextDrawer is now independent.
- cTextDrawer now uses textures to display text on the screen.

- cTextDrawer uses a cache to display textures.
- cTextDrawer's cache is reset every minute.

### DIFF
--- a/src/gui/GuiWindow.cpp
+++ b/src/gui/GuiWindow.cpp
@@ -4,12 +4,12 @@
 #include "drawers/SDLDrawer.hpp"
 
 
-GuiWindow::GuiWindow(const cRectangle &rect, cTextDrawer* _textDrawer ) : GuiObject(rect)
+GuiWindow::GuiWindow(const cRectangle &rect, cTextDrawer* _textDrawer) :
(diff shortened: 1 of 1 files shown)


src/context/ContextCreator.cpp
```diff
@@ -55,15 +55,15 @@ ContextCreator::ContextCreator(SDL_Renderer *renderer, cFileValidator *settingsV
     }  
 
     bene_font = TTF_OpenFont(settingsValidator->getFullName(eGameDirFileName::BENEGESS).c_str(),12);
-    if (game_font == nullptr) {
+    if (bene_font == nullptr) {
```
stefanhendriks, line 58: something for the future, but this feels like we can do via a builder pattern, that looks 'prettier' and pushes out the `settingsValidator` from the creator class.

Mira-Chronos: Step by step :)

PR #721, opened 2025-11-10
Bug / fix particle colorization

Mira-Chronos:
- fig bug of particle colorization
- create texture cache to avoid wasting CPU to create temporary resources
- simplify memory management: --nothing to delete !--

### DIFF
--- a/src/cGame_logic.cpp
+++ b/src/cGame_logic.cpp
@@ -716,6 +716,7 @@ void cGame::shakeScreen(int duration)
 */
 void cGame::shutdown()
 {
+    cParticle::reset();
     cLogger *logger = cLogger::getInstance();
     logger->logHeader("SHUTDOWN");
 
@@ -1228,7 +1229,8 @@ void cGame::think_fading()
 }
 
 cGame::~cGame()
-{}
+{
+}
 
 void cGame::prepareMentatForPlayer()
 {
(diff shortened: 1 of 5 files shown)


src/cGame_logic.cpp
```diff
@@ -716,6 +716,7 @@ void cGame::shakeScreen(int duration)
 */
 void cGame::shutdown()
 {
+    cParticle::reset();
```
stefanhendriks, line 719: We would need to reset cache upon new mission init I think? Because there is no guarentee the same player index have the same colors in the next (skirmish) mission.

Mira-Chronos: Sure! 100% right. GG on you.

Mira-Chronos: We can reset cache at beginning of mission

stefanhendriks: Lets resolve this conversation when we fix this

Mira-Chronos:
This commit [579439d](https://github.com/stefanhendriks/Dune-II---The-Maker/pull/721/commits/579439d7bac9689171cfccabec1e3218d4d1442f) does the job. It will be moved when GameState is operational.
(and i tested the cache with 2 distinct houses)

src/cGame_logic.cpp
```diff
@@ -1169,7 +1170,7 @@ void cGame::setState(int newState)
                         cPlayer &player = players[i];
                         player.evaluateStillAlive();
                     }
-
+                    cParticle::reset();
```
stefanhendriks, line 1173: ah yes, this is a better place as this would - in time - be the `onStateEnter` kind of phase

stefanhendriks (approved):

PR #733, opened 2025-11-12
feature/ display local time & time party duration (Ripps Idea)

Mira-Chronos:
_From Ripps imagination:_

Access is given with touch BACKSLASH

Better explain as visual :

[Capture vidéo du 2025-11-12 20-25-12.webm](https://github.com/user-attachments/assets/e4c9418b-6d2f-4a5f-b3c6-6a629f37008a)

### DIFF
--- a/src/utils/cTimeManager.cpp
+++ b/src/utils/cTimeManager.cpp
@@ -28,9 +28,14 @@ cTimeManager::cTimeManager(cGame *game)
 
 std::string cTimeManager::getCurrentTime() const
 {
-    auto now = std::chrono::system_clock::now();
-    auto now_seconds = std::chrono::time_point_cast<std::chrono::seconds>(now);
-    return std::format("{:%H:%M:%S}", now_seconds);
+    auto now_utc = std::chrono::system_clock::now();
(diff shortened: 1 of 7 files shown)


stefanhendriks:
I tested the feature and noticed it does not render my local time, but rather UTC:

image

My local time (top right) does not match the time at the top-right in D2TM. 

Could this be fixed? Else, we perhaps need to get rid of the local time and just render party time instead.

src/utils/cTimeManager.cpp
```diff
@@ -23,6 +24,13 @@ cTimeManager::cTimeManager(cGame *game)
     durationTime.init(5);
 }
 
+std::string cTimeManager::getCurrentTime() const
+{
+    auto now = std::chrono::system_clock::now();
+    auto now_seconds = std::chrono::time_point_cast<std::chrono::seconds>(now);
```
stefanhendriks: We should approach this a little differently so we use the current timezone as well, so we render the expected local time.

See also: https://stackoverflow.com/a/39274464/214597

Mira-Chronos: New commit should be fix it.

stefanhendriks: Seems to work! 🙌

PR #737, opened 2025-11-13
Add ability to choose techlevel when setting up skirmish map

Mira-Chronos:
Ability to choose TechLevel when choice on setup skirmish map

Capture d’écran du 2025-11-13 22-30-58


stefanhendriks:
I love it when a PR description has an image to show what it did/added/changed! 👍

stefanhendriks (approved):

PR #740, opened 2025-11-15
technical / cTextDrawer improvement with introduction on cTextTextureCache

Mira-Chronos:
On respect on SOLID, split cTextDrawer on cTextDrawer+cTextTextureCache

### DIFF
--- a/src/cGame_logic.cpp
+++ b/src/cGame_logic.cpp
@@ -834,7 +834,7 @@ bool cGame::setupGame()
     ctx->setTextContext(context->createTextContext());
     global_map.setGameContext(ctx.get());
 
-    m_textDrawer = ctx->getTextContext()->gameTextDrawer.get();
+    m_textDrawer = ctx->getTextContext()->getGameTextDrawer();
     //m_textDrawer->setApplyShadow(false);
 
     std::unique_ptr<cSoundPlayer> soundPlayer = std::make_unique<cSoundPlayer>(settingsValidator->getFullName(eGameDirFileName::GFXAUDIO));
--- a/src/context/ContextCreator.cpp
+++ b/src/context/ContextCreator.cpp
@@ -88,11 +88,10 @@ std::unique_ptr<GraphicsContext> ContextCreator::createGraphicsContext()
     return gtx;
 }
 
-std::unique_ptr<TextContext> ContextCreator::createTextContext()
-{
-    auto ttx = std::make_unique<TextContext>();
-    ttx->smallTextDrawer = std::make_unique<cTextDrawer>(small_font);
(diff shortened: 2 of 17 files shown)


src/drawers/cTextTextureCache.cpp
```diff
@@ -0,0 +1,65 @@
...
+    }
+    return it->second.get();
+}
+
+void cTextTextureCache::resetCache() const
+{
+    auto it = m_textCache.begin();
+    while (it != m_textCache.end()) {
+        auto &cacheEntry = it->second;
+        if (cacheEntry->lifeCounter == 0) {
```
stefanhendriks, line 56: I was looking into the usages for `lifeCounter` and see it is only set to `10` at some point, but never decreased. (so what is the 'counting' part?). Then on line 61 below, we set `lifeCounter=0`, but I would expect there it to not matter when resetting cache. Can you revisit this piece if that makes sense?

src/drawers/cTextTextureCache.cpp
```diff
@@ -0,0 +1,65 @@
...
+            if (cacheEntry->texture) SDL_DestroyTexture(cacheEntry->texture);
+            if (cacheEntry->shadowsTexture) SDL_DestroyTexture(cacheEntry->shadowsTexture);
+            it = m_textCache.erase(it);
+        } else {
+            cacheEntry->lifeCounter = 0;
```
stefanhendriks, line 61: for `resetCache` wouldn't we simply remove all entries, regardless of `lifeCounter` ? Just erase all? Also, the removal of an entry is also written in the destructor, so I think we should unify that into one function that takes care of the removal of the textures.

src/drawers/cTextTextureCache.cpp
Mira-Chronos: 10 because 10 > 0
Never decrease because i want to erase all texture not used in the last min.
**So the behavior is:**
- every min, erase all texture in cache with `lifeCounter=0`
- every call in cache give +1 on texture.
Texture never called will no +1 have so she want to be erased after
Why not a boolean ? futur anticipation

src/drawers/cTextTextureCache.cpp
Mira-Chronos: Technically, I'd like to write it (a little) differently, without SDL_Destroy, which I'll do soon using smartpointers. But since it affects the datapacks, I have to do it for the entire code.

src/drawers/cTextTextureCache.cpp
stefanhendriks: I see, the `resetCache` sets it to `0` when the counter was `>0`; so it either gets set to `0` or it will be deleted. If it is used the counter increases.

Issue #741, opened 2025-11-15
Screen shaking when building destroyed


Mira-Chronos:
Re enable screen shaking when building destruction

PR #742, opened 2025-11-15
missing port SDL2 / Enable screen shaking when building is destroyed

Mira-Chronos:
see #741

### DIFF
--- a/src/cGame_logic.cpp
+++ b/src/cGame_logic.cpp
@@ -669,7 +669,7 @@ void cGame::run()
         shakeScreenAndBlitBuffer();
 
         renderDrawer->endDrawingToTexture();
-        renderDrawer->renderSprite(actualRenderer,m_screenShake->getX(),m_screenShake->getY());
+        renderDrawer->renderSprite(actualRenderer, m_screenShake->getX(), m_screenShake->getY());
         SDL_RenderPresent(renderer);
         m_timeManager->waitForCPU(); // wait for CPU to catch up, so we don't run too fast
     }
@@ -1461,8 +1461,7 @@ void cGame::onEventSpecialLaunch(const s_GameEvent &event)
     game.onNotifyGameEvent(newEvent);
 }
 
-void cGame::reduceShaking()
-{
+void cGame::reduceShaking() const {
     m_screenShake->reduce();
 }
 
@@ -2059,11 +2058,6 @@ void cGame::onKeyDownDebugMode(const cKeyboardEvent &event)
             pUnit.die(true, false);
         }
     }
-
-    if (event.hasKey(SDL_SCANCODE_F7)) {
-        m_screenShake->shake(200/3); // fix bug temporary with 3*5ms
-    }
-
 }
 
 void cGame::setMousePosition(int w, int h)
--- a/src/utils/cScreenShake.cpp
+++ b/src/utils/cScreenShake.cpp
@@ -37,11 +37,6 @@ int cScreenShake::getY() const
     return m_shakeY;
 }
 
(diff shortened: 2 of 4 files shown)


src/cGame_logic.cpp
```diff
@@ -573,34 +569,7 @@ void cGame::loadSkirmishMaps() const
...
-            m_shakeX = -abs(offset / 2) + RNG::rnd(offset);
-            m_shakeY = -abs(offset / 2) + RNG::rnd(offset);
-
-            // @Mira recreate shake screen
-        }
-        else {
-            fadeOutOrBlitScreenBuffer();
-        }
-    }
+    m_screenShake->update(m_state, GAME_PLAYING, mapCamera);
```
stefanhendriks, line 572: since we only want to shake screen for `GAME_PLAYING` I would suggest we run this update logic within the `game playing` update loop.  Then we can get rid of the `m_state/GAME_PLAYING` mechanism and remove the 'knowledge of state' in the `m_screenShake` (`cScreenShake`) class altogether. Then it will be agnostic and we can let the screen shake whenever we want eventually.

src/cGame_logic.cpp
```diff
@@ -1497,9 +1463,7 @@ void cGame::onEventSpecialLaunch(const s_GameEvent &event)
 
 void cGame::reduceShaking()
 {
-    if (m_TIMER_shake > 0) {
-        m_TIMER_shake--;
-    }
+    m_screenShake->reduce();
```
stefanhendriks, line 1465: this basically is called from the `think_fast` method. So I suggest we just collapse (remove `reduceShaking`) and then directly call `m_screenShake` from `thinkFast_combat` and rename `reduce` to `thinkFast` as well (for consistency).

src/cGame_logic.cpp
```diff
@@ -2097,8 +2061,7 @@ void cGame::onKeyDownDebugMode(const cKeyboardEvent &event)
     }
 
     if (event.hasKey(SDL_SCANCODE_F7)) {
```
stefanhendriks: This is for debugging purposes and can be removed.

src/utils/cScreenShake.cpp
```diff
@@ -0,0 +1,61 @@
...
+}
+
+void cScreenShake::update(int state, int playingState, cMapCamera *mapCamera)
+{
+    if (m_timer == 0) {
+        m_timer = -1;
+    }
+
+    if (m_timer > 0 && state == playingState) {
+        // the more we get to the 'end' the less we 'throttle'.
```
stefanhendriks, line 47: Seeing this - after a long time - I wonder if we can add another variable that means 'intensity'. Ie, so the shaking is more intens at start and reduces after the fix amount of time. The intensity is now `shakiness` where we do some division and whatnot. Perhaps we can rethink this? In a different PR perhaps.

src/cGame_logic.cpp
Mira-Chronos: Sure ! The entire drawing pipeline will be rewritten when GameState is split.

src/cGame_logic.cpp
Mira-Chronos: Again, it will be completely rewritten when GameState GamePlay will exist.

src/cGame_logic.cpp
Mira-Chronos: why not !

src/utils/cScreenShake.cpp
Mira-Chronos: Sure. Okay for another PR.

PR #745, opened 2025-11-22
technical / step 3: Working gamestate, MentatState creation

Mira-Chronos:
Extract Mentat state  from cGame

Evolution size of files:

file| allegro| sept 2025 | nov 2025 | WinLose | TellHouse | MentatState
| :-- |----:| -----:| -----:| -----:|  -----:|  -----:|
h | 365|396|400|393| 395| 386
cpp| 2226 | 2327 | 2226 |2170 | 2194 | 2094

### DIFF
--- a/src/cGame_logic.cpp
+++ b/src/cGame_logic.cpp
@@ -30,6 +30,7 @@
 #include "gamestates/cSetupSkirmishState.h"
 #include "gamestates/cWinLoseState.h"
 #include "gamestates/cTellHouseState.h"
+#include "gamestates/cMentatState.h"
 #include "ini.h"
 #include "iniDefine.h"
 #include "include/sDataCampaign.h"
@@ -643,12 +644,12 @@ void cGame::drawState()
     }
 
     switch (m_state) {
-        case GAME_BRIEFING:
-        case GAME_LOSEBRIEF:
-        //case GAME_TELLHOUSE: //here
-        case GAME_WINBRIEF:
-            drawStateMentat(m_mentat);
-            break;
+        // case GAME_BRIEFING:
+        // case GAME_LOSEBRIEF:
+        // //case GAME_TELLHOUSE: //here
+        // case GAME_WINBRIEF:
+        //     drawStateMentat(m_mentat);
+        //     break;
         case GAME_PLAYING:
             drawStateCombat();
             break;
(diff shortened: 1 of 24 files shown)


src/cGame_logic.cpp
```diff
@@ -645,12 +630,6 @@ void cGame::drawState()
     }
 
     switch (m_state) {
-        case GAME_BRIEFING:
```
stefanhendriks, line 648: yes awesome!

src/cGame_logic.cpp
```diff
@@ -1253,74 +1244,113 @@ cGame::~cGame()
 
 void cGame::prepareMentatForPlayer()
 {
-    int house = players[HUMAN].getHouse();
     if (m_state == GAME_BRIEFING) {
```
stefanhendriks, line 1247: this sounds like an `onEnter` logic to me. Perhaps `prepareMentatForPlayer` can be removed in favor of that in the future?

src/ini.cpp
```diff
@@ -514,16 +515,18 @@ int getTechLevelByRegion(int iRegion)
 }
 
 
-void cIni::loadScenario(int iHouse, int iRegion, AbstractMentat *pMentat, cReinforcements *reinforcements)
+void cIni::loadScenario(/*int iHouse, int iRegion,*/ AbstractMentat *pMentat, cReinforcements *reinforcements, s_DataCampaign *dataCampaign)
```
stefanhendriks, line 518: I don't see how adding the `s_DataCampaign` helps here, as it introduces coupling here we did not have before. Can we revert that? or is there another reason?

Mira-Chronos: Other reason: end of global variable

Mira-Chronos: Introducing this variable allowed me to simplify the code and eliminate access to cGame. I really want cGame to act as an orchestrator and not as a multimodal platform.

src/cGame_logic.cpp
Mira-Chronos: exactly

src/ini.cpp
```diff
@@ -514,16 +515,18 @@ int getTechLevelByRegion(int iRegion)
...
 {
     game.m_skirmish = false;
     game.missionInit();
+    int iHouse = dataCampaign->housePlayer;
+    int iRegion = dataCampaign->region;
 
     std::string filename = cIniUtils::getScenarioFileName(iHouse, iRegion);
 
-    game.m_mission = getTechLevelByRegion(iRegion);
+    dataCampaign->mission = getTechLevelByRegion(iRegion);
```
stefanhendriks, line 527: ah this is why we need the pointer, as we *set* something here as well 😢

stefanhendriks (approved):

PR #752, opened 2025-11-24
technical / step 4: Working gamestate, CreatorState2

Mira-Chronos:
## Technical PR to use in same place
- `EnumArray` with `enum class`
- `optional`, `unique_ptr` and `EnumArray`

## Goal: 
- Separate the creation of GameStates from their place of use
- Isolate any GameState creation with a cache to avoid recreate them at usage.
- Test modern C++ integration

### DIFF
--- a/src/include/enums.h
+++ b/src/include/enums.h
@@ -348,11 +348,11 @@ enum class SmudgeType : char {
     S_ROCK = 0,
     S_SAND = 1,
     S_WALL = 2,
-    MAX
+    COUNT
 };
 
 enum class GameState : char {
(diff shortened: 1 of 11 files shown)


src/include/enums.h
```diff
@@ -348,5 +320,28 @@ enum class SmudgeType : char {
     S_ROCK = 0,
     S_SAND = 1,
     S_WALL = 2,
-    MAX
+    COUNT
+};
+
+enum class GameState : char {
```
stefanhendriks, line 326: Nice! 🙌

stefanhendriks (approved):

PR #753, opened 2025-11-25
Shai Hulud should never spawn on map border

### LINKED ISSUE #555: Shai-Hulud stuck outside map
Shai Hulud gets stuck outside of the map

[PXL_20220509_071610045 MP](https://user-images.githubusercontent.com/46880246/167361649-adb34196-bf3a-4119-aae3-d40aacba8050.jpg)

see image (sorry, should have scaled it down).

stefanhendriks:
This should fix #555

### DIFF
--- a/src/player/brains/cPlayerBrainSandworm.cpp
+++ b/src/player/brains/cPlayerBrainSandworm.cpp
@@ -38,15 +38,7 @@ void cPlayerBrainSandworm::think()
 
 void cPlayerBrainSandworm::findRandomValidLocationToMoveToAndGoThere(cUnit &pSandWorm) const
 {
-    int placeToMoveTo = -1;
-    for (int iTries = 0; iTries < 5; iTries++) {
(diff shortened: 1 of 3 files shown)


src/player/brains/cPlayerBrainSandworm.cpp
```diff
@@ -38,16 +38,21 @@ void cPlayerBrainSandworm::think()
 
 void cPlayerBrainSandworm::findRandomValidLocationToMoveToAndGoThere(cUnit &pSandWorm) const
 {
-    // find new spot to go to
+    bool foundPlaceToMoveTo = false;
     for (int iTries = 0; iTries < 5; iTries++) {
         int iMoveTo = global_map.getRandomCellWithinMapWithSafeDistanceFromBorder(2);
 
         if (global_map.isCellPassableForWorm(iMoveTo)) {
             pSandWorm.move_to(iMoveTo);
```
stefanhendriks: Extract so that "find place to move to" returns either -1 or a cell. Then there is no need for boolean and it further simplifies logic

PR #765, opened 2025-11-29
  technical / step 5.2: Working on gamestate statistics

Mira-Chronos:
## **Goal**

I use  `s_GameEvent event` to display death of unit or death of struct

For continue, i have two questions:

1) How can we build `GameState Statistics` ? 
The same from Dune2 or more detailled customization ?

2) Do we have to go through the `s_GameEvents` for everything?
Shall we implement more `s_GameEvents` on game like the creation of unity, the total number of units swallowed, trampled...


stefanhendriks:
This PR can be closed, it won't be merged anyway. I already have an idea how I want to implement statistics.

PR #775, opened 2025-12-10
technical / step 6.4: Working gamestate event prepare gameplaying 

Mira-Chronos:
Extract `GamePlaying` from `cGame`

## Changes

Always with the aim of separating the class

- Prepare `cGamePlaying` 

I prepared `cGame` so that I could use `GamePlaying` without any loss of functionality.


src/gamestates/cGamePlaying.cpp
```diff
@@ -74,25 +74,20 @@ void cGamePlaying::thinkSlow()
...
     // for now, call this on game class.
     // TODO: move this "combat" state into own game state class
     drawCombatMouse();
 
-    if (m_drawTime) {
-        auto time = m_timeManager->getCurrentTime();
-        m_textDrawer->drawText(game.m_screenW- cSideBar::SidebarWidth-75, cSideBar::TopBarHeight + 1, Color::white(), time);
-        time = m_timeManager->getCurrentTimer();
-        m_textDrawer->drawText(game.m_screenW- cSideBar::SidebarWidth-75, cSideBar::TopBarHeight + 1+15, Color::white(), time);
+    if (m_game.m_drawTime) {
```
stefanhendriks, line 86: I like the concept of draw fps and such in the state better, it does not make sense to ask `cGame` about this to me

stefanhendriks: Also, what is the difference between `game` and `m_game`? I see it is mixed. I would expect them to be all `m_game`

src/gamestates/cGamePlaying.cpp
```diff
@@ -139,6 +134,7 @@ void cGamePlaying::thinkSlow_stateCombat_evaluatePlayerStatus()
         m_TIMER_evaluatePlayerStatus = 2;
     }
 
+    game.checkMissionWinOrFail();
```
stefanhendriks, line 137: wouldnt it make sense to execute this in the game state?

stefanhendriks (approved):

PR #794, opened 2025-12-22
Support Clang on Mac so that it can build on Mac Intel AND Mac ARM architecture

stefanhendriks:
This PR changes code + CMakeList file so that it builds on the intel mac and arm mac


Mira-Chronos:
This PR work fine on Linux 24.04

PR #829, opened 2026-03-20
technical step 8.3: remove  global var world

Mira-Chronos:
## **Goal**

The 8.X series aims to improve the overall state of the program's global variables and to modify them little by little.

### Changes
- remove  global var `world`

### DIFF
--- a/src/gamestates/cSelectYourNextConquestState.h
+++ b/src/gamestates/cSelectYourNextConquestState.h
@@ -137,4 +137,5 @@ private:
     Graphics* m_gfxworld;
     Graphics* m_gfxinter;
     s_DataCampaign* m_dataCompaign;
+    mutable std::array<cRegion, MAX_REGIONS> world;
 };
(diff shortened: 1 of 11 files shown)


src/gamestates/cSelectYourNextConquestState.h
```diff
@@ -137,4 +137,5 @@ class cSelectYourNextConquestState : public cGameState {
     Graphics* m_gfxworld;
     Graphics* m_gfxinter;
     s_DataCampaign* m_dataCompaign;
+    mutable std::array<cRegion, MAX_REGIONS> world;
```
stefanhendriks, line 140: @Mira-Chronos is this `mutable` keyword needed?

Mira-Chronos: I made it mutable because currently we're modifying things in the `draw` function, which is declared as `const`. I'll also wait until I've rewritten the code for this object more thoroughly before removing the mutable attribute (which I don't like).

stefanhendriks (approved):

PR #835, opened 2026-03-22
technical step 8.9 :  remove global_variables 

Mira-Chronos:
## **Goal**

The 8.X series aims to improve the overall state of the program's global variables and to modify them little by little.

### Changes
- Using cGame to unify the calls to the sub-modules
-  Example : g_Players 
- Example : g_Particles

### DIFF
--- a/src/gameobjects/projectiles/bullet.cpp
+++ b/src/gameobjects/projectiles/bullet.cpp
@@ -646,7 +646,7 @@ s_BulletInfo cBullet::gets_Bullet() const
 
 cPlayer *cBullet::getPlayer() const
 {
-    return &game.getPlayers()[iPlayer];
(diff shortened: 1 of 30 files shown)


src/gameobjects/projectiles/bullet.cpp
```diff
@@ -646,7 +646,7 @@ s_BulletInfo cBullet::gets_Bullet() const
 
 cPlayer *cBullet::getPlayer() const
 {
-    return &g_Players[iPlayer];
+    return &game.getPlayer(iPlayer);
```
stefanhendriks, line 649: even. better would be if this bullet would have a pointer? Makes it way easier?

stefanhendriks (approved):

PR #846, opened 2026-03-26
Render tooltip when pressing `t`

Mira-Chronos:
## **Goal**

Draw ToolTip Reappears.
A gap that deserved to be filled, and now it has been filled with this PR.

### Changes
- move drawToolTip from context to cDrawManager
- activate drawToolTip with key T

## Testing Steps
hit T when mouse over structure

## Screenshots
Capture d’écran du 2026-03-26 18-55-39


stefanhendriks (approved):
works!

PR #851, opened 2026-03-27
Select same type of units on screen when pressing `CTRL-z`

Mira-Chronos:
## **Goal**

select same units type on screen

### Changes
- select from player unit in screen the same unit type.
- link to key `CTRL-z`

## Testing Steps
Select one unit, press `CTRL-z` and the game will select all player's unit in screen from the same type 

## Screenshots
https://github.com/user-attachments/assets/de794d27-f600-4501-9204-37d70b60d08e


stefanhendriks (approved):
works as a charm

PR #891, opened 2026-04-06
Improvement info context on cGame  (Refactoring 1/3)

Mira-Chronos:
## **Goal**

Here is an outline of a new software component that handles the XXXInfos classes of the game.

### Changes
- Create design to change `cGame` by removing all `XXXXINfos` to dedicated class


stefanhendriks (approved):
Looks good!

Mira-Chronos:
Ready for merge. 
`cGame` delegate now all `XXXXInfo` to `cInfoContext`

src/gameobjects/structures/cAbstractStructure.cpp
```diff
@@ -142,26 +142,26 @@ cPlayer *cAbstractStructure::getPlayer()
 int cAbstractStructure::getMaxHP()
 {
     int type = getType();
-    return game.structureInfos[type].hp;
+    return game.m_infoContext->getStructureInfo(type).hp;
```
stefanhendriks, line 145: so for now this is fine, but in the end I think we can even have a direct pointer stored, so we don't need to go via `infoContext`, but we simply store the StructureInfo pointer when we create a structure. Wdyt?

src/gameobjects/structures/cGunTurret.cpp
```diff
@@ -229,7 +229,7 @@ void cGunTurret::think_guard()
 
         int distanceForAttacking = getSight();
         if (lowPower && getType() == RTURRET) {
-            distanceForAttacking = game.structureInfos[TURRET].sight; // HACK HACK: way to reduce distance for rturret on low power
+            distanceForAttacking = game.m_infoContext->getStructureInfo(TURRET).sight; // HACK HACK: way to reduce distance for rturret on low power
```
stefanhendriks, line 232: for example, this could become:

```suggestion
            distanceForAttacking = this->m_structureInfo->sight; // HACK HACK: way to reduce distance for rturret on low power
```

src/map/cMap.h
```diff
@@ -551,7 +551,7 @@ class cMap : public cScenarioObserver {
     std::map<int, int> m_mBloomTimers;
 
     std::shared_ptr<cReinforcements> m_reinforcements;
-    std::shared_ptr<s_TerrainInfo> m_terrainInfo;
+    s_TerrainInfo* m_terrainInfo = nullptr;
```
stefanhendriks, line 554: why a raw pointer here?

stefanhendriks (approved):
see my comments

src/gameobjects/structures/cAbstractStructure.cpp
Mira-Chronos: I am writing a comment: this code is in a transition phase: it is not intended to remain this way.

src/gameobjects/structures/cGunTurret.cpp
Mira-Chronos: No commit please  because it will be rewritten after modification. If I change the pattern, I'll have extra work.

src/map/cMap.h
Mira-Chronos: Because `m_terrainInfo` is now a smart_ptr and `cMap`  does not have to manage the lifetime of `m_terrainInfo`

src/gameobjects/structures/cGunTurret.cpp
stefanhendriks: that is fine, but what do you think about the idea?

Mira-Chronos: As say to the pdf file, this is a transitory code, and to respect the Demeter rule, it must be rewritten as you mentioned.

PR #903, opened 2026-04-07
Improvement coding : removing define and transform it as constexpr if possible

Mira-Chronos:
## **Goal**

Remove some old define  to class and improve code quality

### Changes
- Fix `ASTAR temp_map` as static to avoid recreate it on every call
- Remove define from `definitions.h` to right destination
- Transform define to constexpr as possible

### DIFF
--- a/src/gameobjects/units/cPathFinder.cpp
+++ b/src/gameobjects/units/cPathFinder.cpp
@@ -8,15 +8,12 @@
 #include "map/MapGeometry.hpp"
 #include "utils/d2tm_math.h"
 
+// Initialize the static temp_map
+ASTAR cPathFinder::temp_map[16384];
+
 // Path creation definitions / var
 #define CLOSED        -1
 #define OPEN          0
-
-struct ASTAR {
-    int cost;
-    int parent;
-    int state;
-};
 
 /*
   Pathfinder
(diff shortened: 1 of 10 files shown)


src/gameobjects/units/cPathFinder.cpp
```diff
@@ -8,16 +8,13 @@
 #include "map/MapGeometry.hpp"
 #include "utils/d2tm_math.h"
 
+// Initialize the static temp_map
+ASTAR cPathFinder::temp_map[16384];
```
stefanhendriks, line 12: We should comment/make clear there is a relationship with the map size here

src/gameobjects/units/cPathFinder.cpp
```diff
@@ -43,8 +40,6 @@ struct ASTAR {
   */
 int cPathFinder::createPath(int iUnitId, int iPathCountUnits)
 {
-    ASTAR temp_map[16384]; // 4096 = 64x64 map, 16384 = 128x128 map
```
stefanhendriks, line 46: This comment

src/gameobjects/units/cPathFinder.h
```diff
@@ -2,10 +2,16 @@
...
+};
+
 class cPathFinder {
 public:
     static int createPath(int iUnitId, int iPathCountUnits);
     static int returnCloseGoal(int iCll, int iMyCell, int iID);
-private: 
-
+private:
+    static ASTAR temp_map[16384];
```
stefanhendriks, line 16: Defined twice?

stefanhendriks (approved):
Some comments, in general good change. Less define is better. Did not test functionality

src/gameobjects/units/cPathFinder.h
Mira-Chronos: Define once, create once !

src/gameobjects/units/cPathFinder.cpp
Mira-Chronos: Yet, but not yet !

PR #904, opened 2026-04-08
Technical / remove renderer from `cPlayer` class

Mira-Chronos:
## **Goal**

`cPlayer` didn't need access to renderer yet.

### Changes
- Create `texture_utils.h` 
- Move `createTextureFromIndexedSurfaceWithPalette` to `texture_utils.h` 

## Testing Steps
All units from houses have the correct color

### DIFF
--- a/src/utils/texture_utils.cpp
+++ b/src/utils/texture_utils.cpp
@@ -1,0 +1,13 @@
+#include "texture_utils.h"
+#include "player/cPlayer.h"
+#include "drawers/SDLDrawer.hpp"
+#include "game/cGame.h"
+#include "include/d2tmc.h"
+
+Texture* createPlayerTextureFromIndexedSurfaceWithPalette(cPlayer* player, SDL_Surface* referenceSurface, int paletteIndexForTransparency) {
+    assert(player && "player must be given");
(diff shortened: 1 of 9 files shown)


src/utils/texture_utils.cpp
```diff
@@ -0,0 +1,13 @@
+#include "texture_utils.h"
+#include "player/cPlayer.h"
+#include "drawers/SDLDrawer.hpp"
+#include "game/cGame.h"
+#include "include/d2tmc.h"
+
+Texture* createPlayerTextureFromIndexedSurfaceWithPalette(cPlayer* player, SDL_Surface* referenceSurface, int paletteIndexForTransparency) {
```
stefanhendriks, line 7: Do you think we will need it elsewhere then player? It felt correct to be at player level. Although, this basically moves the implementation outside of the player class, you still *need* it. If you really want to detach it from player I guess this implementation should be done differently, where we get the `swapColor` from the player and pass it as argument in this function. Wdyt?

Mira-Chronos: Play didn't know about rendering so this function is only for rendering so i think it didn't be i the right place

stefanhendriks (approved):

Issue #914, opened 2026-04-12
[Bug] Skirmish game with extra players spawning in 1v1 match.


Ripps1:
If in skirmish setup screen, you select a four player map, assign computers to each player and then select a 2 player map.
the setup screen looks fine (2 players)
but the game will start with all four players but since there is only 2 spawn locations strange things happen

[Video here](https://faaf.me/82c5bc57)

PR #916, opened 2026-04-13
Fix bug #914 : remove CPU players if new map has less starting points than currently set

Mira-Chronos:
For #914

## **Goal**
remove CPU players if new map has less starting points than currently set

### DIFF
--- a/src/gamestates/cSetupSkirmishState.cpp
+++ b/src/gamestates/cSetupSkirmishState.cpp
@@ -30,6 +30,7 @@
 #include <algorithm>
 #include <utility>
 #include <cassert>
+#include <iostream>
 
 #include "config.h"
 
@@ -576,13 +577,15 @@ void cSetupSkirmishState::prepareSkirmishGameToPlayAndTransitionToCombatState(in
 
     /* set up starting positions */
     std::vector<int> iStartPositions;
-
+    std::cout << "number of start cell in prepareSkirmishGameToPlayAndTransitionToCombatState: " << iStartingPoints << std::endl;
     int startCellsOnSkirmishMap = 0;
     for (int s = 0; s < 5; s++) {
         int startPosition = selectedMap.iStartCell[s];
-        if (startPosition < 0) continue;
+        if (startPosition < 1) continue;
+        std::cout << "Adding start position: " << startPosition << std::endl;
         iStartPositions.push_back(startPosition);
     }
+
 
     startCellsOnSkirmishMap = iStartPositions.size();
 
@@ -1135,6 +1138,11 @@ void cSetupSkirmishState::onMouseLeftButtonClickedAtMapList(const cRectangle &se
                     if (s > -1) {
                         iStartingPoints++;
                     }
+                    std::cout << "number of start cell: " << iStartingPoints << std::endl;
+                }
+                // remove CPU players if new map has less starting points than currently set
+                for (int p = iStartingPoints; p < (AI_WORM - 1); p++) {
+                    skirmishPlayer[p].bPlaying = false;
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


src/gamestates/cSetupSkirmishState.cpp
```diff
@@ -576,14 +577,16 @@ void cSetupSkirmishState::prepareSkirmishGameToPlayAndTransitionToCombatState(in
 
     /* set up starting positions */
     std::vector<int> iStartPositions;
-
+    std::cout << "number of start cell in prepareSkirmishGameToPlayAndTransitionToCombatState: " << iStartingPoints << std::endl;
     int startCellsOnSkirmishMap = 0;
     for (int s = 0; s < 5; s++) {
         int startPosition = selectedMap.iStartCell[s];
-        if (startPosition < 0) continue;
+        if (startPosition < 1) continue;
```
stefanhendriks: are you sure? This should return a cell, and cell `0` is valid (although probably out of bounds for a unit to be spawned)

src/gamestates/cSetupSkirmishState.cpp
```diff
@@ -1135,6 +1138,11 @@ void cSetupSkirmishState::onMouseLeftButtonClickedAtMapList(const cRectangle &se
                     if (s > -1) {
                         iStartingPoints++;
                     }
+                    std::cout << "number of start cell: " << iStartingPoints << std::endl;
+                }
+                // remove CPU players if new map has less starting points than currently set
+                for (int p = iStartingPoints; p < (AI_WORM - 1); p++) {
```
stefanhendriks, line 1140: this is the actual fix right?

src/gamestates/cSetupSkirmishState.cpp
```diff
@@ -576,14 +577,16 @@ void cSetupSkirmishState::prepareSkirmishGameToPlayAndTransitionToCombatState(in
...
     /* set up starting positions */
     std::vector<int> iStartPositions;
-
+    std::cout << "number of start cell in prepareSkirmishGameToPlayAndTransitionToCombatState: " << iStartingPoints << std::endl;
     int startCellsOnSkirmishMap = 0;
     for (int s = 0; s < 5; s++) {
         int startPosition = selectedMap.iStartCell[s];
-        if (startPosition < 0) continue;
+        if (startPosition < 1) continue;
+        std::cout << "Adding start position: " << startPosition << std::endl;
```
stefanhendriks: we should get rid of the `cout` statements

src/gamestates/cSetupSkirmishState.cpp
Mira-Chronos: oups !! That was a mistake on my part. It's wrong

src/gamestates/cSetupSkirmishState.cpp
Mira-Chronos: That was a mistake on my part.

src/gamestates/cSetupSkirmishState.cpp
Mira-Chronos: Yes, it's the solution

src/gamestates/cSetupSkirmishState.cpp
Mira-Chronos: I fixed in [f6f8dc3](https://github.com/stefanhendriks/Dune-II---The-Maker/pull/916/commits/f6f8dc3ff23ebfa023ae522ea0f82b3c9bb77db6)

src/gamestates/cSetupSkirmishState.cpp
Mira-Chronos: I fixed in [f6f8dc3](https://github.com/stefanhendriks/Dune-II---The-Maker/pull/916/commits/f6f8dc3ff23ebfa023ae522ea0f82b3c9bb77db6)

PR #925, opened 2026-04-15
Creating GameInterface (1)

Mira-Chronos:
## **Goal**

A technical way to mask `cGame` to the project between his Interface  `cGameInterface`

I create a test with `cGame` and gameState `cChooseHouseState`

With this interface, no rebuild and better lambda function on class how need to call cGame


src/game/cGameInterface.cpp
```diff
@@ -0,0 +1,82 @@
...
+cMouse* cGameInterface::getMouse() const
+{
+    return m_igame->getMouse();
+}
+
+void cGameInterface::prepareMentatToTellAboutHouse(int house) const
+{
+    switch(house) {
+    case ATREIDES:
+            m_igame->prepareMentatToTellAboutHouse(ATREIDES);
```
stefanhendriks, line 26: tbh, I would expect these things in the mentat state, not in another class

stefanhendriks: Ie, this would be on a `onEnter` for a state

Mira-Chronos: ---> #947

PR #946, opened 2026-04-18
Using find_package with Catch2

Mira-Chronos:
## **Goal**

The installation of catch2 is handled by the operating system


stefanhendriks:
This PR breaks on ARM64 for Mac, because of some incompatibility between Clang and GCC for Catch2 (when preinstalled using brew). Will try to get the Mac version to compile and build via Clang uniformly first, and then retry this PR.

PR #962, opened 2026-04-19
fix hard-coded values on cGame

### LINKED ISSUE #954: [Bug] Game does not respect CameraEdgeMove=false in settings.ini
Game does not respect the CameraEdgeMove=false setting in the settings.ini file.

I believe this issue has been around since the SDL2 migration. cant be sure, but it effects all the installs i currently have going back to the start of the year. (I delete installs for older versions to save space (i retain the download))

Mira-Chronos:
Close #954 

## **Goal**

Access to the camera parameters via the `setting.ini` file is now correctly read.
```C++
    m_cameraDragMoveSpeed
    m_cameraBorderOrKeyMoveSpeed
    m_cameraEdgeMove
 ```
get values from settings

### Changes
- Move hard-coded values on constructor

### DIFF
--- a/src/game/cGame_logic.cpp
+++ b/src/game/cGame_logic.cpp
@@ -127,6 +127,10 @@ cGame::cGame()
     m_mapCamera = nullptr;
     m_drawManager = nullptr;
 
+    m_cameraDragMoveSpeed=0.5f;
+    m_cameraBorderOrKeyMoveSpeed=0.5;
+    m_cameraEdgeMove = true;
+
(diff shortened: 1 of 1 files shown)


stefanhendriks:
So the init methode would override the settings ini?

stefanhendriks (approved):

Mira-Chronos:
> So the init methode would override the settings ini?

yes

stefanhendriks:
verified, works!

PR #965, opened 2026-04-19
Test bugs ripps NEVER MERGE (under penalty of having an explosive log)

Mira-Chronos:
Special for @Ripps1 

## **Goal**

More details on console


Mira-Chronos:
Time out.

PR #968, opened 2026-04-19
technical move initPlayer (3)

Mira-Chronos:
## **Goal**

Move function to this right place


stefanhendriks:
When switching to mission 9, I get a crash:

```
std::unique_ptr::get[abi:ne180100]() const unique_ptr.h:247
cMentatState::loadScenario(cReinforcements *) cMentatState.cpp:126
cGame::loadScenario() cGame_logic.cpp:1017
cGameInterface::loadScenario() const cGameInterface.cpp:173
cSelectYourNextConquestState::loadScenarioAndTransitionToNextState(int) cSelectYourNextConquestState.cpp:420
cSelectYourNextConquestState::onMouseLeftButtonClicked(const s_MouseEvent &) cSelectYourNextConquestState.cpp:679
cSelectYourNextConquestState::onNotifyMouseEvent(const s_MouseEvent &) cSelectYourNextConquestState.cpp:648
cGame::onNotifyMouseEvent(const s_MouseEvent &) cGame_logic.cpp:1268
cInteractionManager::onNotifyMouseEvent(const s_MouseEvent &) cInteractionManager.cpp:107
cMouse::updateState() cMouse.cpp:148
cGame::updateMouseAndKeyboardState() cGame_logic.cpp:357
cGame::run() cGame_logic.cpp:444
main main.cpp:55
```
asking claude to help with this.

stefanhendriks:
According to claude this is the cause:

```
 Root cause: cGame::loadScenario() at line 1016 does dynamic_cast<cMentatState *>(m_states[GAME_BRIEFING]). When accessed via cSelectMissionState (the dev mission-select shortcut), the game jumps directly to
  GAME_REGION without ever visiting GAME_BRIEFING, so m_states[GAME_BRIEFING] is null. dynamic_cast returns null, and calling pState->loadScenario() on it causes the unique_ptr::get() crash at line 126 (accessing
  m_mentat via a null this).

  There's also a secondary bug at line 101: memset(m_states, 0, sizeof(cGameState *)) only zeroes 8 bytes (one pointer), leaving m_states[1..19] uninitialized.

  The loadScenario() call in loadScenarioAndTransitionToNextState is actually redundant — the new cMentatState created in transitionStateIfRequired already calls cIni::loadScenario in its prepareMentat constructor
  chain. The fix is to guard against the null case and fix the memset.

```

I did use the select mission , so it got that right.

stefanhendriks:
The mission select from Skirmish is broken. I noticed, if you play a campaign, lets say with House Harkonnen. Then jump to Skirmish and play as Atreides, but from there open Options->Select mission. You will be selecting a mission as Harkonnen (and not as Atreides you were playing).

stefanhendriks:
Bug found was there before this PR, so agreed to merged. Create issue for bug https://github.com/stefanhendriks/Dune-II---The-Maker/issues/995

PR #972, opened 2026-04-19
Fix renderer texture (NOT TO MERGE)

Mira-Chronos:
For LINK_TO_GH_ISSUE here

## **Goal**
using OpenGL on windows to avoid glitches


Mira-Chronos:
wrong patch

PR #976, opened 2026-04-20
Technical players pointers (5)

Mira-Chronos:
## **Goal**

`cPlayers` used mix of `Objet&` and `Objet*`
Yet `cPlayers` use only `Objet*`

We can now create Player with parameters !!!!!!!!!!!


Mira-Chronos:
Wrong commit

PR #982, opened 2026-04-20
Technical cplayers using services (7) 

Mira-Chronos:
## **Goal**

Clean `cPlayers` and `cPlayer` to be modern

### Changes
- rip `cGame` call
- better modern code

### DIFF
--- a/src/game/cGameInterface.cpp
+++ b/src/game/cGameInterface.cpp
@@ -6,6 +6,7 @@
 #include "controls/cMouse.h"
 #include "gameobjects/units/cReinforcements.h"
 #include "include/sGameEvent.h"
+#include "utils/cStructureUtils.h"
 
 #include <format>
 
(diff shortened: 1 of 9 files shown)


src/game/cGame_logic.cpp
```diff
@@ -697,6 +698,9 @@ bool cGame::setupGame()
     // I need m_renderDrawer to create cPreviewMaps
     m_PreviewMaps = std::make_shared<cPreviewMaps>(m_renderDrawer, m_gameSettings->m_debugMode);
 
+    // Injection of services
+    m_players->serviceInit(m_services.get());
```
stefanhendriks, line 702: we can't inject this via constructor?

src/game/cGameInterface.cpp
```diff
@@ -193,6 +194,41 @@ bool cGameInterface::playMusicByType(int iType) const
...
+}
+
+int cGameInterface::getTotalPowerUsageForPlayer(cPlayer *pPlayer) const
+{
+    return m_igame->m_structureUtils->getTotalPowerUsageForPlayer(pPlayer);
+}
+
+int cGameInterface::getTotalPowerOutForPlayer(cPlayer *pPlayer) const
+{
+    return m_igame->m_structureUtils->getTotalPowerOutForPlayer(pPlayer);
```
stefanhendriks, line 219: this feels so weird, seeing it like that. Why this is not a property from player? Not something you can change now, but architectural wise, it feels backwards.

stefanhendriks (commented):
Some small questions/remarks.

src/game/cGame_logic.cpp
Mira-Chronos: No, i tested and impossible until i have game as global variable

src/game/cGameInterface.cpp
Mira-Chronos: I jsut need more time to fix it (on 85 files)

PR #984, opened 2026-04-20
Fix graphical glitches

### LINKED ISSUE #625: Artifact in select you next conquest
Image

Slicing on picture

### LINKED ISSUE #815: [Bug] Broken Flag animations on certain buildings
no description needed when a picture says a thousand words ;-)

Image

yea to be honest a few more words won't hurt.

it seems to effect mostly (if not only) buildings that use 2 flags.

Mira-Chronos:
From issues #815 and #625 

## **Goal**

Fix glitches on graphical picture on Windows system.

close #815 
close #625

### DIFF
--- a/src/drawers/SDLDrawer.cpp
+++ b/src/drawers/SDLDrawer.cpp
@@ -334,7 +334,15 @@ Texture *SDLDrawer::createTextureFromIndexedSurfaceWithPalette(SDL_Surface *refe
     // Copy pixels
     SDL_LockSurface(referenceSurface);
     SDL_LockSurface(modifiableSurface);
-    memcpy(modifiableSurface->pixels, referenceSurface->pixels, referenceSurface->h * referenceSurface->pitch);
+//    memcpy(modifiableSurface->pixels, referenceSurface->pixels, referenceSurface->h * referenceSurface->pitch);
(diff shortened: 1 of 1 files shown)


stefanhendriks:
Can you tell me what caused the bug? I see you copy data differently now, but I am unsure why this needs to be in a for loop for example. Can you try to explain a bit?

stefanhendriks:
Claude explains:

```
 The fix works because of a pitch mismatch between the source and destination surfaces.

  The bug: memcpy copies the entire pixel buffer in one shot using referenceSurface->pitch as the row stride. On Windows, SDL may allocate the destination surface (modifiableSurface) with a different pitch (row byte
   width, including padding) than the source. Blindly copying h * referenceSurface->pitch bytes would write source row data into the wrong positions of the destination rows.

  The fix: Copies row-by-row, using each surface's own pitch as the stride offset, but copying only referenceSurface->w bytes (1 byte per pixel for an 8-bit indexed surface) per row. This correctly handles the case
  where src->pitch != dst->pitch.

  The pitch difference is likely Windows-specific due to how SDL allocates textures aligned to GPU/driver requirements on that platform, which explains the windows-only label.
  ```

PR #992, opened 2026-04-22
Units will respond to move order even if mouse is outside map boundary

Mira-Chronos:
## **Goal**

A unit moves around the battlefield regardless of where you click on the camera view, even if that click is outside the map. The unit will move to the nearest cell to the click, using Manhattan distance.

## Screenshots (optional)

https://github.com/user-attachments/assets/82084a92-0d64-47f0-97f0-bfc38bd997bb

### DIFF
--- a/src/controls/cGameControlsContext.cpp
+++ b/src/controls/cGameControlsContext.cpp
@@ -25,6 +25,7 @@ cGameControlsContext::cGameControlsContext(cPlayer *player, cMouse *mouse) :
     m_mouseHoveringOverUnitId(-1),
     m_mouseOnBattleField(false),
     m_mouseCell(-99),
+    m_mouseCellClamped(0),
     m_player(player),
     m_state(MOUSESTATE_SELECT),
     m_prevState(MOUSESTATE_SELECT),
@@ -70,14 +71,20 @@ void cGameControlsContext::updateMouseCell(const cPoint &coords)
         return;
     }
     m_mouseOnBattleField = true;
-    m_mouseCell = getMouseCellFromScreen(coords.x, coords.y);
+    m_mouseCell = getMouseCellFromScreen(coords.x, coords.y, false);
+    if (m_mouseCell<0) { // clampe mouse cell to map
+        m_mouseCellClamped = getMouseCellFromScreen(coords.x, coords.y, true);
+    }
 }
 
-int cGameControlsContext::getMouseCellFromScreen(int mouseX, int mouseY) const
+int cGameControlsContext::getMouseCellFromScreen(int mouseX, int mouseY, bool clamped) const
 {
     int absMapX = game.m_mapCamera->getAbsMapMouseX(mouseX);
     int absMapY = game.m_mapCamera->getAbsMapMouseY(mouseY);
-    return game.m_mapCamera->getCellFromAbsolutePosition(absMapX, absMapY);
+    if (clamped)
+        return game.m_mapCamera->getCellClampedFromAbsolutePosition(absMapX, absMapY);
+    else
+        return game.m_mapCamera->getCellFromAbsolutePosition(absMapX, absMapY);
 }
 
 void cGameControlsContext::determineHoveringOverStructureId()
--- a/src/controls/cGameControlsContext.cpp
+++ b/src/controls/cGameControlsContext.cpp
@@ -71,21 +71,24 @@ void cGameControlsContext::updateMouseCell(const cPoint &coords)
         return;
     }
     m_mouseOnBattleField = true;
-    m_mouseCell = getMouseCellFromScreen(coords.x, coords.y, false);
-    if (m_mouseCell<0) { // clampe mouse cell to map
-        m_mouseCellClamped = getMouseCellFromScreen(coords.x, coords.y, true);
+    m_mouseCell = getMouseCellFromScreen(coords.x, coords.y);
+    if (m_mouseCell < 0) { // clamp mouse cell to map
+        m_mouseCellClamped = getMouseCellFromScreenClamped(coords.x, coords.y);
     }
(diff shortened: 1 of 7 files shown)


src/controls/cGameControlsContext.cpp
```diff
@@ -70,14 +71,20 @@ void cGameControlsContext::updateMouseCell(const cPoint &coords)
...
-int cGameControlsContext::getMouseCellFromScreen(int mouseX, int mouseY) const
+int cGameControlsContext::getMouseCellFromScreen(int mouseX, int mouseY, bool clamped) const
 {
     int absMapX = game.m_mapCamera->getAbsMapMouseX(mouseX);
     int absMapY = game.m_mapCamera->getAbsMapMouseY(mouseY);
-    return game.m_mapCamera->getCellFromAbsolutePosition(absMapX, absMapY);
+    if (clamped)
+        return game.m_mapCamera->getCellClampedFromAbsolutePosition(absMapX, absMapY);
+    else
+        return game.m_mapCamera->getCellFromAbsolutePosition(absMapX, absMapY);
```
stefanhendriks:
```suggestion
    if (clamped) {
        return game.m_mapCamera->getCellClampedFromAbsolutePosition(absMapX, absMapY);
    } else {
        return game.m_mapCamera->getCellFromAbsolutePosition(absMapX, absMapY);
    }
```

Lets try to use always `{}` to wrap code.

src/controls/cGameControlsContext.cpp
```diff
@@ -70,14 +71,20 @@ void cGameControlsContext::updateMouseCell(const cPoint &coords)
         return;
     }
     m_mouseOnBattleField = true;
-    m_mouseCell = getMouseCellFromScreen(coords.x, coords.y);
+    m_mouseCell = getMouseCellFromScreen(coords.x, coords.y, false);
+    if (m_mouseCell<0) { // clampe mouse cell to map
+        m_mouseCellClamped = getMouseCellFromScreen(coords.x, coords.y, true);
```
stefanhendriks: Lets create 2 functions instead using a flag. Better practice.

Ie:


```suggestion
        m_mouseCellClamped = getMouseCellFromScreenClamped(coords.x, coords.y);
```

src/controls/cGameControlsContext.cpp
```diff
@@ -70,14 +71,20 @@ void cGameControlsContext::updateMouseCell(const cPoint &coords)
...
 }
 
-int cGameControlsContext::getMouseCellFromScreen(int mouseX, int mouseY) const
+int cGameControlsContext::getMouseCellFromScreen(int mouseX, int mouseY, bool clamped) const
 {
     int absMapX = game.m_mapCamera->getAbsMapMouseX(mouseX);
     int absMapY = game.m_mapCamera->getAbsMapMouseY(mouseY);
-    return game.m_mapCamera->getCellFromAbsolutePosition(absMapX, absMapY);
+    if (clamped)
+        return game.m_mapCamera->getCellClampedFromAbsolutePosition(absMapX, absMapY);
```
stefanhendriks: this is no longer needed if having 2 functions. 

final nit: The naming of the function `getCellClampedFromAbsolutePosition` can be to `getCellFromAbsolutePositionClamped`

So the structure/wording is the same (consistency wins)

src/map/cMapCamera.cpp
```diff
@@ -179,6 +179,11 @@ int cMapCamera::getCellFromAbsolutePosition(int x, int y)
     return m_mapGeometry->getCellWithMapDimensions((x / 32), (y / 32));
 }
 
+int cMapCamera::getCellClampedFromAbsolutePosition(int x, int y)
+{
+    return m_mapGeometry->getCellClampedWithMapDimensions((x / 32), (y / 32));
```
stefanhendriks:
```suggestion
    return m_mapGeometry->getCellWithMapDimensionsClamped((x / 32), (y / 32));
```

src/map/MapGeometry.cpp
```diff
@@ -137,6 +137,15 @@ int MapGeometry::getCellWithMapDimensions(int x, int y) const
     return (y * mapWidth) + x;
 }
 
+int MapGeometry::getCellClampedWithMapDimensions(int x, int y) const
+{
+    if (x < 1) x= 1;
```
stefanhendriks: This looks like `fixCoordinatesToBeWithinPlayableMap`.

From a point we can compute cell. This function looks like duplication to me now.

stefanhendriks (commented):
Few comments, I think we can simplify the fix.

src/controls/cGameControlsContext.cpp
Mira-Chronos: Imagine, that's what I did at first. And it was simply copy and paste, which bothered me because all the work goes into the final function; everything else is pipeline.

src/map/MapGeometry.cpp
Mira-Chronos: Oh no absolutely not !!!! If all is clamped, the soft explose because -1 is a value that has signification. We cannot clamp for all

stefanhendriks:
```
cPoint MapGeometry::fixCoordinatesToBeWithinPlayableMap(int x, int y) const
{
    return {std::clamp(x, 1, mapWidth - 2), std::clamp(y, 1, mapHeight - 2)};
}
```
this function makes sure there will be no -1, so you can safely do that.

src/controls/cGameControlsContext.cpp
stefanhendriks: Well, having a switch to influence a functions behaviour [is usually not desired](https://martinfowler.com/bliki/FlagArgument.html).

PR #996, opened 2026-04-23
Allow sending units via minimap

### LINKED ISSUE #291: Send units to destination via minimap
From #287 

It should be possible to send selected units to destination via minimap. 

- [ ] only do this when units are selected
- [ ] you should not navigate minimap when units are selected
- [ ] mouse cursor should change into a small 'move to' cursor

Mira-Chronos:
From #291 

## **Goal**

Send units to destination via minimap is now possible with this patch

### Changes
- `cUnits` can move units
- `cMinimapDrawer` change behavior if units are selected

## Testing Steps
I created maps with markers to check the PR

## Screenshots

https://github.com/user-attachments/assets/53cee2fd-bbf8-4006-af5d-1ef303e88e4e

close #291

### DIFF
--- a/src/drawers/cMiniMapDrawer.cpp
+++ b/src/drawers/cMiniMapDrawer.cpp
@@ -213,6 +213,9 @@ void cMiniMapDrawer::onNotifyMouseEvent(const s_MouseEvent &event)
         case eMouseEventType::MOUSE_LEFT_BUTTON_PRESSED:
             onMousePressedLeft(event);
             return;
+        case eMouseEventType::MOUSE_RIGHT_BUTTON_PRESSED:
+            onMousePressedRight(event);
+            return;
         default:
             return;
     }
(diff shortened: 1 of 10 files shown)


src/gameobjects/units/cUnits.cpp
```diff
@@ -213,3 +213,23 @@ int cUnits::unitCreate(int iCll, int unitType, int iPlayer, bool bOnStart, bool
 // {
 //     return unitCreate(iCll, unitType, iPlayer, bOnStart, isReinforcement, 1.0f);
 // }
+
+bool cUnits::areUnitsSelected() const
+{
+    for (const auto& unit : m_units) {
```
stefanhendriks: this will not work, as we need to know for which player.

src/drawers/cMiniMapDrawer.cpp
```diff
@@ -392,7 +394,20 @@ void cMiniMapDrawer::onMousePressedLeft(const s_MouseEvent &event)
     if (m_RectFullMinimap.isPointWithin(event.coords.x, event.coords.y) && // on minimap space
         !game.getMouse()->isBoxSelecting() // pressed the mouse and not boxing anything..
     ) {
-
+        if (game.m_gameObjectsContext->getUnits()->areUnitsSelected() == true) {
```
stefanhendriks: we should query the units I believe via the player, not global state.

stefanhendriks (commented):
As discussed, I'll process the feedback

PR #998, opened 2026-04-23
Improvement deploy unit event (14)

Mira-Chronos:
## **Goal**

Using `Events` to create and deploy unit on battlefield.
Can deploy units without having `cUnits` link to did it.

### DIFF
--- a/src/game/cGame_logic.cpp
+++ b/src/game/cGame_logic.cpp
@@ -1098,6 +1098,7 @@ void cGame::onNotifyGameEvent(const s_GameEvent &event)
         case eGameEventType::GAME_EVENT_DESTROYED:
             onEventEntityDestroyed(event);
             break;
+        case eGameEventType::GAME_EVENT_DEPLOY_UNIT:
         case eGameEventType::GAME_EVENT_CREATE_UNIT:
             onEventDeployUnit(event);
             break;
--- a/src/game/cGame_logic.cpp
+++ b/src/game/cGame_logic.cpp
@@ -1161,7 +1161,11 @@ void cGame::onEventDeployUnit(const s_GameEvent &event) {
         true,
         event.isReinforce
     );
-    cLogger::getInstance()->log(LOG_INFO, COMP_GAME, "Deploy unit", std::format("Deployed unit of type {} at cell {} for player {}, id={}", event.entitySpecificType, event.atCell, event.player->getId(), id));
+    if (id < 0) {
+        cLogger::getInstance()->log(LOG_ERROR, COMP_GAME, "Deploy unit", std::format("Failed to deploy unit of type {} at cell {} for player {}", event.entitySpecificType, event.atCell, event.player->getId()));
+    } else {
+        cLogger::getInstance()->log(LOG_INFO, COMP_GAME, "Deploy unit", std::format("Successfully deployed unit of type {} at cell {} for player {}, id={}", event.entitySpecificType, event.atCell, event.player->getId(), id));
+    }
 }
 
 void cGame::onEventSpecialLaunch(const s_GameEvent &event) const {
--- a/src/game/cGame_logic.cpp
+++ b/src/game/cGame_logic.cpp
@@ -1099,6 +1099,7 @@ void cGame::onNotifyGameEvent(const s_GameEvent &event)
             onEventEntityDestroyed(event);
             break;
         case eGameEventType::GAME_EVENT_DEPLOY_UNIT:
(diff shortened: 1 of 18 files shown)


src/gamestates/cSetupSkirmishState.cpp
```diff
@@ -838,7 +847,15 @@ void cSetupSkirmishState::prepareSkirmishGameToPlayAndTransitionToCombatState(in
                 continue;
             }
             logbook(std::format("Spawning sandworm at {}", cell));
-            cUnits::unitCreate(cell, SANDWORM, AI_WORM, true);
+            // cUnits::unitCreate(cell, SANDWORM, AI_WORM, true);
+            s_GameEvent event {
+                .eventType = eGameEventType::GAME_EVENT_DEPLOY_UNIT,
```
stefanhendriks: if the intent is to create a unit, lets call it `GAME_EVENT_CREATE_UNIT`

Mira-Chronos: There were no `GAME_EVENT_CREATE_UNIT`  and  `GAME_EVENT_DEPLOY_UNIT`  seem to do the same thing. Or i'm wrong ?

stefanhendriks: `GAME_EVENT_DEPLOY_UNIT` is spawned when a unit is done being repaired, and thus has to leave the structure. 

Best would be to introduce a new event. Although, I am not sure if I like the approach of spawning events to produce a unit. I did not mind having to use `cUnit::unitCreate`. What is the benefit?

Mira-Chronos: Simply simplify the current code because `unite::create` is static and won't stay that way. Hence the need to change it. Furthermore, what could be more practical than using the `GameEvent` which is designed for this purpose?

Mira-Chronos: I can create a new EVENT named `GAME_EVENT_CREATE_UNIT` as you suggested.

stefanhendriks: yes, we should do that. We cannot use DEPLOY event

src/game/cGame_logic.cpp
```diff
@@ -1098,6 +1098,10 @@ void cGame::onNotifyGameEvent(const s_GameEvent &event)
         case eGameEventType::GAME_EVENT_DESTROYED:
             onEventEntityDestroyed(event);
             break;
+        case eGameEventType::GAME_EVENT_DEPLOY_UNIT:
```
stefanhendriks:
```suggestion
```
Only create upon CREATE_UNIT, as `GAME_EVENT_DEPLOY_UNIT` triggers different logic in `cRepairFacility`.

Mira-Chronos: after: 
```C++
if (iNewUnitType > -1) {
                    int id = cUnits::unitCreate(position.iCell, iNewUnitType, iPlayer, true, isReinforcement);

                    if (id > -1) {
                        m_map->cellSetIdForLayer(position.iCell, MAPID_UNITS, id);
                    }
                }
```
the case `id>1` is redundant, already included in `unitCreate`

yet:

```C++
                if (iNewUnitType > -1) {
                    s_GameEvent event {
                        .eventType = eGameEventType::GAME_EVENT_DEPLOY_UNIT,
                        .entityType = eBuildType::UNIT,
                        .entityID = iID,
                        .player = getPlayer(),
                        .entitySpecificType = iNewUnitType,
                        .atCell = position.iCell,
                        .isReinforce = isReinforcement,
                        .originId = iID,
                        .originType = eBuildType::UNIT,
                    };
                    m_interface->onNotifyGameEvent(event);
                }
```
and after refactoring we have :
```C++
                if (iNewUnitType > -1) {
                    const s_GameEvent event {
                        .eventType = eGameEventType::GAME_EVENT_DEPLOY_UNIT,
                        .data = DeployUnitEvent {
                            .iCell = position.iCell,
                            .unitType = iNewUnitType,
                            .iPlayer = getPlayer()->getId(),
                            .bOnStart = true,
                            .isReinforcement = isReinforcement
                        }
                    };
                    m_interface->onNotifyGameEvent(event);
                }
```

Mira-Chronos: Argggg GAME_EVENT_DEPLOY_UNIT is used another part on the game ... for the same action. I rip this duplicate code

Mira-Chronos: There you go, it's clarified and relayed.

PR #999, opened 2026-04-23
Improvement modern gameevent 1 (15)

Mira-Chronos:
## **Goal**

Beginning the process of splitting `sGameEvent` to eventually use only the `std::variant data`

### DIFF
--- a/src/controls/mousestates/cMouseDeployState.cpp
+++ b/src/controls/mousestates/cMouseDeployState.cpp
@@ -56,15 +56,15 @@ void cMouseDeployState::onMouseLeftButtonClicked()
     cBuildingListItem *itemToDeploy = m_player->getSideBar()->getList(eListType::LIST_PALACE)->getItemToDeploy();
     if (itemToDeploy == nullptr) return;
 
-    s_GameEvent event {
+    // launch missile by sending event
+    const s_GameEvent event {
         .eventType = eGameEventType::GAME_EVENT_SPECIAL_LAUNCH,
-        .entityType = eBuildType::SPECIAL,
-        .entityID = -1,
-        .player = m_player,
-        .entitySpecificType = itemToDeploy->getBuildId(),
-        .atCell = mouseCell,
-        .isReinforce = false,
-        .buildingListItem = itemToDeploy
+        .data = LaunchDeathHandEvent {
+            .targetCell = mouseCell,
+            .itemToLaunch = itemToDeploy,
+            .player = m_player,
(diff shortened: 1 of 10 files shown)


stefanhendriks:
## Code review notes (analysis by Claude Code, requested by @stefanhendriks)

### What this PR does (in Java terms)

`std::variant` is roughly equivalent to a **sealed interface + pattern matching** (Java 21). The two new structs (`DeployUnitEvent`, `LaunchDeathHandEvent`) are the equivalent of **records** — typed, named data bags per event type. `std::monostate` is the "no data" case, similar to a null/empty permitted type.

The `std::get_if` dispatch in `cGame_logic.cpp:1096` and `:1104` is equivalent to Java's `instanceof` pattern matching:
```java
if (event.data instanceof DeployUnitEvent deployEvent) { ... }
```

### Potential pitfalls

**1. Silent no-ops on `get_if` failure (`cGame_logic.cpp:1096`, `1104`)**
If `get_if` returns `nullptr` (wrong type in `.data`, or `monostate` because a caller forgot to populate it), the `if` body is silently skipped — no error, no log, nothing happens. The event type says "launch Death Hand" but the handler never runs.

Suggested fix: add an `assert` or log for the else-branch:
```cpp
case eGameEventType::GAME_EVENT_SPECIAL_LAUNCH:
    if (const auto *launchEvent = std::get_if<LaunchDeathHandEvent>(&event.data)) {
        onEventSpecialLaunch(*launchEvent);
    } else {
        assert(false && "GAME_EVENT_SPECIAL_LAUNCH requires LaunchDeathHandEvent in data");
    }
    break;
```
This is especially easy to hit because the old flat fields (`entityType`, `atCell`, etc.) still exist on `s_GameEvent`, which could lull a future caller into populating those instead of `.data`.

**2. `GAME_EVENT_DEPLOY_UNIT` may be a dead path**
In `cGamePlaying.cpp` one call site was changed to emit `GAME_EVENT_DEPLOY_UNIT`, but the switch in `cGame_logic.cpp:1103` only handles `GAME_EVENT_CREATE_UNIT`. If `GAME_EVENT_DEPLOY_UNIT` isn't handled elsewhere, that event silently falls through to `default` and does nothing. Worth verifying.

**3. No exhaustiveness checking**
Unlike Java sealed classes + switch expressions, adding a new type to the variant does not force all `get_if` sites to be updated — the compiler won't warn. `std::visit` with a visitor struct would give exhaustiveness, but at the cost of verbosity. Something to keep in mind as more event types are migrated.

PR #1005, opened 2026-04-24
Show move cursor when hovering minimap with units selected

### LINKED ISSUE #1004: When mouse hovers over minimap, render a smaller mouse move icon to indicate units will sent to move there
An improvement over https://github.com/stefanhendriks/Dune-II---The-Maker/pull/996

When mouse hovers over minimap, then give a visual clue that the move action is 'enabled'. Ie, instead of a normal mouse pointer show the move icon. 

If no units are selected, show the normal mouse pointer.

stefanhendriks:
## Summary

- When the mouse hovers over the minimap and units are selected, the cursor changes to the move icon to indicate units will be sent there on click
- When no units are selected, the normal cursor is shown
- Cursor is re-asserted after click/press events to prevent the mouse state machine from overwriting it

## How it works

`cMiniMapDrawer::updateMouseCursor()` centralises the cursor logic. It is called from `onMouseAt` (hover) and also at the end of `onNotifyMouseEvent` for click/press events. This is necessary because `cMouseUnitsSelectedState` calls `setTile()` before the minimap drawer runs in the observer chain, so the minimap must re-assert the cursor last.

## Testing Steps

1. Select one or more units
2. Hover the mouse over the minimap — cursor should change to the move icon
3. Click to move units — cursor should remain the move icon (not revert to normal)
4. Deselect all units, hover over minimap — cursor should show the normal pointer

Closes #1004

image

### DIFF
--- a/src/drawers/cMiniMapDrawer.cpp
+++ b/src/drawers/cMiniMapDrawer.cpp
@@ -209,17 +209,18 @@ void cMiniMapDrawer::onNotifyMouseEvent(const s_MouseEvent &event)
             return;
         case eMouseEventType::MOUSE_LEFT_BUTTON_CLICKED:
             onMouseClickedLeft(event);
-            return;
+            break;
         case eMouseEventType::MOUSE_LEFT_BUTTON_PRESSED:
             onMousePressedLeft(event);
-            return;
+            break;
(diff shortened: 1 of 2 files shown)


Ripps1:
Units given orders this way do not respond verbally to the command (no "yes sir" sound)
Cursor doesn't update for me either.

will check more tommorow

PR #1076, opened 2026-05-03
Improvement facing

Mira-Chronos:
# **Goal**

simply a refactoring with class enums and without defining the faces of a unit

### Changes

See the code, this PR was created for that

### DIFF
--- a/src/include/Facing.h
+++ b/src/include/Facing.h
@@ -1,0 +1,32 @@
+#pragma once
+
+#include "utils/RNG.hpp"
+
+enum class Facing : int {
+    UP = 0,
+    UPRIGHT = 1,
+    RIGHT = 2,
+    DOWNRIGHT = 3,
+    DOWN = 4,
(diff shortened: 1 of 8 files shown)


src/include/Facing.h
```diff
@@ -0,0 +1,32 @@
...
+    const int count = facingToInt(Facing::COUNT);
+    int normalized = value % count;
+    if (normalized < 0) {
+        normalized += count;
+    }
+    return static_cast<Facing>(normalized);
+}
+
+inline Facing randomFacing() {
+    return facingFromInt(RNG::rnd(facingToInt(Facing::COUNT)));
```
stefanhendriks, line 31: at some point, this should become a flexible thing. So no need for hard-coded values or enum values. It should be depending on the unit-info bitmap where we can compute the amount of angles based on the amount of images (horizontally) in a sprite.

stefanhendriks (approved):
Very nice improvement! 👍

PR #1098, opened 2026-05-04
Skirmish screen: truncate map names longer than 18 chars

### LINKED ISSUE #1093: Skirmish Screen: map names extend beyond available area
Image

we should shorten map names when they are too long to display.

stefanhendriks:
## Summary
- Fixes #1093
- Map names that exceed the available pixel width of the map button (137px) are now truncated with \`...\` in the skirmish map selection grid
- Truncation is based on rendered pixel width via \`getTextLength()\`, so proportional fonts are handled correctly
- No string allocation occurs when the name already fits

## Test plan
- [ ] Open skirmish screen and verify map names that are too wide are truncated with \`...\`
- [ ] Verify short/narrow map names are unchanged
- [ ] Verify selected/hovered state colors still apply to truncated names

### DIFF
--- a/src/gamestates/cSetupSkirmishState.cpp
+++ b/src/gamestates/cSetupSkirmishState.cpp
@@ -1229,11 +1229,18 @@ void cSetupSkirmishState::drawMapList(const cRectangle &selectMapArea) const
             textColor = colorDisabled;
         }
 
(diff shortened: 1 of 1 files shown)


stefanhendriks:
tested and works

image

PR #1115, opened 2026-05-07
Improvement pathfinding

Mira-Chronos:
## **Goal**

Rewriting pathfinding code to be more clear and readable

### Changes

- Rewrite some minus part on the class
- Remove all comment 
- Explicit all methods 
- Using waypoints (blue)
Capture d’écran du 2026-05-08 08-32-50
- No recreating path when path already created
- Fix zig zag

Capture d’écran du 2026-05-08 08-29-31
- Fix wrong first waypoint
Capture d’écran du 2026-05-08 08-28-39


src/gameobjects/units/cPathFinder.cpp
```diff
@@ -57,48 +56,45 @@ void cPathFinder::serviceInit(sGameServices* services)
...
+        return -99;
     }
 
-    m_activeUnit = m_objects->getUnit(iUnitId);
+    m_activeUnit = m_objects->getUnit(unitId);
     if (m_activeUnit == nullptr) {
-        logbook("CREATE_PATH -- END (no unit iUnitId found)");
-        return -99; // Wut!?
+        logbook("CREATE_PATH -- END (no unit unitId found)");
+        return -99;
```
stefanhendriks, line 69: we should consider returning an enum instead, more clear than magic numbers.

src/gameobjects/units/cPathFinder.cpp
```diff
@@ -382,268 +333,605 @@ void cPathFinder::executeCreatePathSearch()
...
         if (pathCell > -1) {
             if (m_mapGeometry->isCellAdjacentToOtherCell(m_activeUnit->getCell(), pathCell)) {
-                m_activeUnit->movement.iPathIndex = i;
+                m_activeUnit->movement.iPathIndex = pathIndex;
+            }
+        }
+    }
+}
+
+int cPathFinder::createPath(int unitId, int pathCountUnitsBudget)
```
stefanhendriks, line 514: we should consider returning a result object here, status of path creation + a potential path (if result is OK).

src/gameobjects/units/cPathFinder.cpp
```diff
@@ -382,268 +333,605 @@ void cPathFinder::executeCreatePathSearch()
...
+{
+    // Full pipeline: validate, search, rebuild, then apply to unit movement.
+    int validationStatus = validateCreatePathInput(unitId);
+    if (validationStatus != 0) {
+        return validationStatus;
+    }
+
+    initializeCreatePathSearch(pathCountUnitsBudget);
+    executeCreatePathSearch();
```
stefanhendriks, line 523: why not return succes here? No need for `m_success` then?

src/gameobjects/units/cPathFinder.cpp
Mira-Chronos: Nice idea. It's add to issues.

src/gameobjects/units/cPathFinder.cpp
Mira-Chronos: Since the PathFinder class has access to the unit it is processing, it modifies the variables of its path in real time.

src/gameobjects/units/cPathFinder.cpp
Mira-Chronos: because there is no return other than success from this place

PR #1125, opened 2026-05-10
Improvement pathfinding fields

Mira-Chronos:
## **Goal**

Using fields to find a one-way path without backtrace

### Changes
- Compute from destination to source
- way that unit can see use line without fields
- no full diagonal way

## Screenshots
Capture d’écran du 2026-05-10 11-14-56

### DIFF
--- a/src/gameobjects/units/cUnit.cpp
+++ b/src/gameobjects/units/cUnit.cpp
@@ -2362,9 +2362,21 @@ void cUnit::think_attack_sandworm()
         return;
     }
 
+    // Sandworms may only chase/eat units on worm-passable terrain.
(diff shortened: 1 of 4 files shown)


src/gameobjects/units/cUnit.cpp
```diff
@@ -3589,6 +3595,7 @@ void cUnit::thinkFast_guard_sandworm()
         if (potentialDinner->getPlayer()->isSameTeamAs(getPlayer())) continue;
         if (potentialDinner->isAirbornUnit()) continue;
         if (potentialDinner->isSandworm()) continue; // don't eat other worms
+        if (!m_map->isCellPassableForWorm(potentialDinner->position.iCell)) continue;
```
stefanhendriks, line 3647: smart!

Issue #1132, opened 2026-05-13
Multi-select rectangle and viewport drag line is difficult to see


Mira-Chronos:
It's difficult to see the line because sometimes the gray is too dark. The idea is brilliant, but can we limit the spread of dark gray?

PR #1134, opened 2026-05-13
Multi-select box and drag line are more visible

stefanhendriks:
## Summary

- The pulsing white multi-select rectangle and viewport drag line could fade as dark as (80,80,80), making them hard to see against terrain (issue #1132)
- Added `getColorFadeSelectedLimited(Color color, float minFade)` to `cScreenFader` (and threaded through `cGame` / `cGameInterface`) which clamps the fade multiplier to a caller-supplied minimum
- `getColorFadeSelected` now delegates to the internal `int r,g,b` implementation (with `minFade = 0.0f`), so the clamping logic lives in one place — no behaviour change for existing callers
- The `int r,g,b` overload is private in `cScreenFader` (internal use only); the public API only exposes the `Color` overload
- `drawCombatMouse` uses `getColorFadeSelectedLimited(Color::White, 0.5f)`: the two visuals now pulse between (128,128,128) and (255,255,255) instead of (80,80,80)–(255,255,255)

## Test plan

- [ ] Build with `./rebuildmacos.sh` — all tests pass
- [ ] Run `./create_release.sh` and launch the game
- [ ] Start a skirmish and drag a multi-select box — the pulsing rectangle should remain clearly visible at its darkest point
- [ ] Hold right-click to scroll the viewport — the drag line should similarly remain visible
- [ ] Confirm selected-unit highlight boxes and sidebar item fading are visually unchanged

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/game/cGame.h
+++ b/src/game/cGame.h
@@ -145,7 +145,6 @@ public:
 
     int getMaxVolume();
 
-    Color getColorFadeSelectedLimited(int r, int g, int b, float minFade);
     Color getColorFadeSelectedLimited(Color color, float minFade);
     Color getColorFadeSelected(int r, int g, int b, bool rFlag = true, bool gFlag = true, bool bFlag = true);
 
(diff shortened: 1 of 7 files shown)


Ripps1:
The Box should always start out with its white colour, and THEN begin the fade from white to gray.

PR #1161, opened 2026-05-15
Print stacktrace when assertion fails on platforms that support it

Mira-Chronos:
## **Goal**

Assert get more information when condition are wrong

### Changes
- Using customized Assert
- Assert use stacktrace feature funtion & line
- Assert create dump.log to easy information

## Screenshots
image

### DIFF
--- a/src/building/cItemBuilder.cpp
+++ b/src/building/cItemBuilder.cpp
@@ -24,7 +24,7 @@
 #include "context/cGameObjectContext.h"
 #include "game/cGameSettings.h"
 
-#include <cassert>
+#include "include/cAssert.h"
 
 #include "data/gfxaudio.h"
 
@@ -33,8 +33,8 @@ cItemBuilder::cItemBuilder(cPlayer *thePlayer, cBuildingListUpdater *buildingLis
     m_buildingListUpdater(buildingListUpdater),
     m_buildItemMultiplierEnabled(false)
 {
-    assert(thePlayer!=nullptr);
-    assert(buildingListUpdater!=nullptr);
+    my_assert(thePlayer!=nullptr);
+    my_assert(buildingListUpdater!=nullptr);
     removeAllItems();
     memset(m_timers, 0, sizeof(m_timers));
 }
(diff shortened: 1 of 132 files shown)


src/building/cItemBuilder.cpp
```diff
@@ -33,8 +33,8 @@ cItemBuilder::cItemBuilder(cPlayer *thePlayer, cBuildingListUpdater *buildingLis
     m_buildingListUpdater(buildingListUpdater),
     m_buildItemMultiplierEnabled(false)
 {
-    assert(thePlayer!=nullptr);
-    assert(buildingListUpdater!=nullptr);
+    my_assert(thePlayer!=nullptr);
```
stefanhendriks: Can we perhaps call this something like `d2tm_assert` ? I find `my_assert` a bit too silly 😆

stefanhendriks:
Make sure to change the `.yml` files where `-DCMAKE_BUILD_TYPE=Release` becomes `-DCMAKE_BUILD_TYPE=RelWithDebInfo`

So that the debug symbols are known for the bleeding edge releases and such.

stefanhendriks:
We need to make this with a switch so it works with Linux, but also compiles (bot not work) for other platforms.

Stacktrace is not yet usable for Mac nor Windows, afaik -> https://en.cppreference.com/cpp/compiler_support/23

image

We can circumvent this by adding `#ifdefs` I suppose

src/include/cAssert.h
```diff
@@ -23,12 +25,8 @@ inline void d2tm_assert(
     report << "Function: " << location.function_name() << "\n";
     report << "----------------------------------------\n";
     report << "Call Stack:\n";
-    
-    // Stack C++23 capture 
+#ifdef __cpp_lib_stacktrace
     auto trace = std::stacktrace::current();
-    
-    // Index 0: std::stacktrace::current() // rip !
-    // Index 1: my_my_assert()
```
Mira-Chronos: it was to explain why we start at index 1 not 0

PR #1183, opened 2026-05-17
Bug repeat defeated message

Mira-Chronos:
## **Goal**

Avoid:
repete_event

Removal of multiple player calls from logs and visual notifications

### DIFF
--- a/src/gameobjects/players/cPlayer.cpp
+++ b/src/gameobjects/players/cPlayer.cpp
@@ -1617,6 +1617,9 @@ void cPlayer::onNotifyGameEvent(const s_GameEvent &event)
     switch (event.eventType) {
         case eGameEventType::GAME_EVENT_PLAYER_DEFEATED:
             if (const auto *commonEvent = std::get_if<CommonEvent>(&event.data)) {
+                if (commonEvent->player != this) {
+                    break;
+                }
                 auto msg = std::format("Player {} ({}) has been defeated.", commonEvent->player->getId(),
                                        commonEvent->player->getHouseName());
                 // addNotification(msg, eNotificationType::BAD);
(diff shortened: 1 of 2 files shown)


src/gameobjects/players/cPlayer.cpp
```diff
@@ -1617,6 +1617,9 @@ void cPlayer::onNotifyGameEvent(const s_GameEvent &event)
     switch (event.eventType) {
         case eGameEventType::GAME_EVENT_PLAYER_DEFEATED:
             if (const auto *commonEvent = std::get_if<CommonEvent>(&event.data)) {
+                if (commonEvent->player != this) {
```
stefanhendriks, line 1622: Why is this the fix?

When a player is defeated, it should be broadcasted to everyone. Including to the player who was defeated.

src/gamestates/cGamePlaying.cpp
```diff
@@ -320,7 +321,7 @@ void cGamePlaying::evaluatePlayerStatus()
             }
             // TODO: event : Player joined/became alive, etc?
         }
-        m_TIMER_evaluatePlayerStatus = 2;
+        m_TIMER_evaluatePlayerStatus = 5;
```
stefanhendriks, line 324: Why change it to 5? This won't fix it either.

src/gameobjects/players/cPlayer.cpp
stefanhendriks: I think what you mean is to only show notifications to the player applicable. *there* we could do a filter. I think at the notification-rendering place we should filter on player. If we change player (`m_controllingPlayer`) then we will see different notifications (applicable to that player).

src/gamestates/cGamePlaying.cpp
Mira-Chronos: why evaluate so many times per second ?

src/gameobjects/players/cPlayer.cpp
Mira-Chronos: Okay, i will fix it for the notification. 
it wasn't unclear on my mind that notification will be used for AI because it's create only logs.

Mira-Chronos: The problem doesn't stem from the `cGamePlaying.cpp:305` loop itself. This method correctly emits a single defeat event. The real issue is that this event is then broadcast to all players in `cGame_logic.cpp:1134`, and each instance of `cPlayer.cpp:1618` recreates the same global notification. As a result, a single defeat generates multiple identical notifications.

Issue #1038, opened 2026-04-30
New shortcut keys (Viewport Stuff) (part 3)


Ripps1:
[Part 1](https://github.com/stefanhendriks/Dune-II---The-Maker/issues/864)
[Part 2](https://github.com/stefanhendriks/Dune-II---The-Maker/issues/895)

Misc. Controls:

Since some control conflicts have arisen, let me suggest a few more keyboard controls, this time relating specifically to the Zoom/Viewport

Insert = Focus Camera on last message (e.g harvester lost, enemy unit approaching etc.)
Delete = ( I have no good ideas for this key )
Home = Focus Camera on Construction Yard
End = Reset Zoom Level
Page up = Zoom out (Alternative to mouse control)
Page down = Zoom in (Alternative to mouse control)

As for the delete key, for the sake of making every button on that part of the keyboard do _something_
we could do something like use it for the sell/retire structures/units feature that may or may not come, cycle camera position through unit group locations, [bookmarked map locations](https://github.com/stefanhendriks/Dune-II---The-Maker/issues/1039) etc.

[Part 4](https://github.com/stefanhendriks/Dune-II---The-Maker/issues/1039)

stefanhendriks:
Love `insert` key, is interesting, because then these messages/events need more information, but it can be done.

Home, kind of like `h` I suppose? Maybe a 'primary' `CONSTYARD`?

I am still looking for some kind of navigate the sidebar with keys. Perhaps with VIM like keys? hjkl?

Ripps1:
> Love `insert` key, is interesting, because then these messages/events need more information, but it can be done.
> 
> Home, kind of like `h` I suppose? Maybe a 'primary' `CONSTYARD`?
> 
> I am still looking for some kind of navigate the sidebar with keys. Perhaps with VIM like keys? hjkl?

For navigating the sidebar i had ideas too.
Check them out [here](https://github.com/stefanhendriks/Dune-II---The-Maker/issues/895#issuecomment-4355718839)

PR #1189, opened 2026-05-17
Updated settings.ini to avoid command conflicts

Ripps1:
Remapped some keys to avoid conflicts

[Issue #1038](https://github.com/stefanhendriks/Dune-II---The-Maker/issues/1038)

## **Goal**

I remapped `CENTER_ON_HOME`, `ZOOM_RESET`, `EDITOR_ZOOM_IN` & `EDITOR_ZOOM_OUT` to avoid conflicts (the zoom thing is a personal preference i admit, but i feel it makes more intuitive sense)

### Changes

Updated default settings.ini included with the game.


resources/bin/settings.ini
```diff
@@ -87,9 +87,9 @@ SCROLL_LEFT=Left
 SCROLL_RIGHT=Right
 SCROLL_UP=Up
 SCROLL_DOWN=Down
-CENTER_ON_HOME=H
+CENTER_ON_HOME=HOME
```
stefanhendriks, line 90: I loved the H key tbf. Home is hard to use on my macbook ;-)

stefanhendriks: What does this conflict with btw? perhaps we can re-wire the other one.

Ripps1: possible to assign the function to both keys?

Ripps1: > What does this conflict with btw? perhaps we can re-wire the other one.

Z (zoom) conflicts with select all units of the same type
H (focus on conyard) conflicts with select all harvesters.

The zoom keys had no conflicts but i liked my change better.

stefanhendriks: I thought that was bound to CTRL-Z and CTRL-H right?

Afaik just pressing H would send me home, and pressing Z would restore zoom level (not select units).

stefanhendriks:
What did we conclude here?

resources/bin/settings.ini
stefanhendriks: Ah, if I remember correctly, the problem is . When pressing Z you restore zoom level. But if you press CTRL-Z (select all units of the same type) it will trigger the action for CTRL-Z but *also* restore zoom level.

Hence, the "just Z" keybinding should *not* trigger when CTRL is pressed. This is a bug I think.

So yeah, basically you what you said lol

stefanhendriks:
With recent merge we no longer have clashing keybindings. I suggest we close this pr

Issue #1209, opened 2026-05-19
Infantry are too weak and have strange values


Ripps1:
So after watching my brother play a few times and discussing with him about how infantry are weak, why they were weak in the original game and why they are weak in this game.

So I started my own investigation into the game rules (`game.ini`)

I noticed some odd stats;
So these are the stats for the single soldier, if we assume a squad is simply 3 soldiers then...
```
; INFANTRY / TROOPERS / ETC

; -----------------
; Unit: Soldier
; -----------------
[UNIT: SOLDIER]
; Unit properties
HitPoints=40
MoveSpeed=5
TurnSpeed=1
FireTwice=false
AttackFrequency=10
Sight=2
Range=1

; Build properties + Sidebar
BuildTime=12
Cost=60
```

...these stats fall into question.

```
; -----------------
; Unit: Infantry
; -----------------
[UNIT: INFANTRY]
; Unit properties
HitPoints=80
MoveSpeed=5
TurnSpeed=1
FireTwice=true
AttackFrequency=13
Sight=3
Range=2

; Build properties + Sidebar
BuildTime=20
Cost=100
```

To me:
- the attack frequency should be the same
- HitPoints should be 120 (3x40)
- Build time should be 36
- Cost should be 180

So i started looking at other infantry too.

```
; -----------------
; Unit: Trooper
; -----------------
[UNIT: TROOPER]
; Unit properties
HitPoints=60
MoveSpeed=6
TurnSpeed=1
FireTwice=false
AttackFrequency=15
Sight=3
Range=3

; Build properties + Sidebar
BuildTime=28
Cost=80

; -----------------
; Unit: Troopers
; -----------------
[UNIT: TROOPERS]
; Unit properties
HitPoints=100
MoveSpeed=6
TurnSpeed=1
FireTwice=true
AttackFrequency=18
Sight=4
Range=4

; Build properties + Sidebar
BuildTime=38
Cost=175
```
The Troopers suffer from similar strange stats, like why are the trooper squads arguably worse than the singletons?

```
; -----------------
; Unit: Single fremen (Atreides)
; the buildtime/cost etc is not used here!
; -----------------
[UNIT: ONEFREMEN]
; Unit properties
HitPoints=120
MoveSpeed=4
TurnSpeed=2
AttackFrequency=16
Sight=6
Range=5

; changing build time or cost here will not do anything
; this is stored in a separate entry, a so called "Special". These are not yet
; supported in the game.ini file yet
[...]

PR #1210, opened 2026-05-19
Updates to game rules (game.ini) in an effort re-balance infantry

Ripps1:
These suggested improvements are based on the fact many infantry had strange stats that didn't reflect what we saw (for example being a squad doesn't have 3x health as you would expect for having 3x units wrapped into one)

[Issue #1209](https://github.com/stefanhendriks/Dune-II---The-Maker/issues/1209)

## **Goal**

Balance changes to infantry to help make them less weak and better reflect what you actually see on the battefield
(a trooper squad should have 3x the health of a single trooper etc.)

### Changes

I made changes to the game.ini
Increased health of squads (3x of the singleton versions)
Increased rate of fire for infantry & squads.
Changed range values to squads and singletons share the same value.

## Testing Steps
Let me do my thing.

## Additional Notes

These changes should be considered experimental and could be reverted or altered further if testing demands it.

### DIFF
--- a/resources/game.ini
+++ b/resources/game.ini
@@ -265,7 +265,7 @@ Cost=125
 HitPoints=40
 MoveSpeed=1
 TurnSpeed=1
-AttackFrequency=10
+AttackFrequency=6
 Sight=2
 Range=2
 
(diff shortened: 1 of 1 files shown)


Ripps1:
I now realize why its x2 and not x3 - please ignore the changes regarding health.

PR #1243, opened 2026-06-02
Migrate to SDL3

### LINKED ISSUE #1241: Migrate from SDL2 to SDL3
Currently we run against SDL2.

SDL3 has been out for a while; this ticket is about migrating to that version.

This allows for better input handling and making d2tm possible on other devices! 🎉

stefanhendriks:
Closes #1241

## Summary

- Migrates the entire engine from SDL2 to SDL3 (3.4.x) plus SDL3_image, SDL3_ttf, and SDL3_mixer
- The game compiles cleanly on macOS (Apple Silicon, gcc-15)
- The game runs: main menu, music, skirmish, sound effects, screenshots, and fullscreen all work

## Key changes by area

**Build system** — CMakeLists.txt and install scripts updated for SDL3. Ubuntu CI builds SDL3 from source (not yet in Ubuntu 24.04 repos). Windows CI uses MSYS2 `mingw-w64-x86_64-SDL3/image/ttf/mixer` packages.

**I/O layer** — `SDL_RWops` fully replaced by `SDL_IOStream` throughout `cPack` and `cDataPack`; read/write calls updated for SDL3's byte-count semantics.

**Audio** — SDL3_mixer 3.x ships a completely new `MIX_*` API (the old `Mix_*` channel model is gone). `cSoundPlayer` now owns a `MIX_Mixer` and a dedicated `MIX_Track` for music; sound effects use `MIX_PlayAudio` (fire-and-forget); `DataPack` was unified: `Mix_Music*`/`Mix_Chunk*` merged into `MIX_Audio*`. Null-mixer guard added so the game runs cleanly on headless systems with no audio device.

**Window/renderer** — `SDL_CreateWindow` no longer takes x/y; HiDPI is automatic; `SDL_CreateRenderer` drops the driver index; VSync set via `SDL_SetRenderVSync`; logical size via `SDL_SetRenderLogicalPresentation`.

**Rendering** — All draw functions renamed (`SDL_RenderCopy` → `SDL_RenderTexture`, `SDL_RenderDrawRect` → `SDL_RenderRect`, etc.); destination rects changed from `SDL_Rect` to `SDL_FRect`; `cRectangle::toSDLF()` added alongside `toSDL()`. Per-texture scale mode set to `SDL_SCALEMODE_NEAREST` everywhere to keep pixel art crisp (SDL3 defaults to bilinear). `SDL_SetDefaultTextureScaleMode(NEAREST)` applied on the renderer so the logical presentation's internal texture also uses nearest-neighbour.

**Surface/pixel format** — `SDL_Surface::format` is now an enum value, not a struct pointer; all palette/pixel-format field accesses updated (`SDL_BITSPERPIXEL`, `SDL_GetSurfacePalette`, `SDL_GetPixelFormatDetails`); `SDL_FreeSurface` → `SDL_DestroySurface` everywhere.

**Events** — All event constants renamed to `SDL_EVENT_*` style; window focus events split into individual `SDL_EVENT_WINDOW_*` types; `cFocusManager` refactored; `event.key.keysym.scancode` → `event.key.scancode`; mouse coordinates cast from float.

**TTF** — `TTF_RenderUTF8_Blended` → `TTF_RenderText_Solid` (fully-opaque pixels scale cleanly with nearest-neighbour; Blended's partial-alpha edges look hazy at non-integer upscale factors); `TTF_SizeUTF8` → `TTF_GetStringSize`; `TTF_FontHeight` → `TTF_GetFontHeight`; `TTF_GetError` → `SDL_GetError`.

**Fullscreen / HiDPI** — `SDL_LOGICAL_PRESENTATION_INTEGER_SCALE` used in fullscreen so every game pixel maps to an equal-sized block on screen. On displays where the integer factor would be < 2 (e.g. 1080p with a 1365×768 logical frame), falls back to `SDL_LOGICAL_PRESENTATION_LETTERBOX` to avoid tiny game windows. Mouse events mapped through `SDL_RenderCoordinatesFromWindow` to correctly reach screen edges. `SDL_VIDEO_MAC_FULLSCREEN_SPACES=0` set on macOS; `SDL_SyncWindow` called after entering fullscreen to guarantee renderer output size is final before logical presentation is applied.

**Shutdown** — `SDLTextureDeleter` guarded against post-`SDL_Quit` destruction. Explicit `Graphics`/`Texture` reset in `cGame::shutdown()` ensures shared_ptr references are released while the renderer is still alive. `SDL_Quit` guarded with `SDL_WasInit(0)` to prevent double-shutdown crash on Linux (SDL3 registers an atexit handler).

**Build tool** — `rebuildmacos.sh` now copies the freshly built binary into `bin/` automatically.

## Test plan

- [x] `./install_dependencies_macosx.sh` on macOS (Apple Silicon)
- [x] `./rebuildmacos.sh` — compiles with no errors
- [x] `./create_release.sh` then `./bin/d2tm` — main menu appears
- [x] Music plays in the main menu
- [x] Start a skirmish game — units animate, sound effects play
- [x] Press screenshot key — file saved in `screen_*/` folder
- [x] Toggle fullscreen (F11) — no crash, pixel-art renders crisp

### DIFF
--- a/.github/workflows/build_manual.yml
+++ b/.github/workflows/build_manual.yml
@@ -60,7 +60,7 @@ jobs:
         with:
           msystem: MINGW64 # D2TM is built within a MinGW64 bit environment
           update: true
-          install: git gzip zip unzip mingw-w64-x86_64-toolchain mingw64/mingw-w64-x86_64-SDL2 mingw64/mingw-w64-x86_64-SDL2_mixer mingw64/mingw-w64-x86_64-SDL2_image mingw64/mingw-w64-x86_64-SDL2_ttf mingw64/mingw-w64-x86_64-cmake mingw-w64-x86_64-catch
+          install: git gzip zip unzip mingw-w64-x86_64-toolchain mingw-w64-x86_64-sdl3 mingw-w64-x86_64-sdl3-mixer mingw-w64-x86_64-sdl3-image mingw-w64-x86_64-sdl3-ttf mingw64/mingw-w64-x86_64-cmake mingw-w64-x86_64-catch
       - name: configure Pagefile
         uses: al-cheb/configure-pagefile-action@v1.5
         with:
--- a/.github/workflows/build_master.yml
+++ b/.github/workflows/build_master.yml
@@ -22,7 +22,7 @@ jobs:
         with:
           msystem: MINGW64 # D2TM is built within a MinGW64 bit environment
           update: true
-          install: git gzip zip unzip mingw-w64-x86_64-toolchain mingw64/mingw-w64-x86_64-SDL2 mingw64/mingw-w64-x86_64-SDL2_mixer mingw64/mingw-w64-x86_64-SDL2_image mingw64/mingw-w64-x86_64-SDL2_ttf mingw64/mingw-w64-x86_64-cmake mingw-w64-x86_64-catch
+          install: git gzip zip unzip mingw-w64-x86_64-toolchain mingw-w64-x86_64-sdl3 mingw-w64-x86_64-sdl3-mixer mingw-w64-x86_64-sdl3-image mingw-w64-x86_64-sdl3-ttf mingw64/mingw-w64-x86_64-cmake mingw-w64-x86_64-catch
       - name: configure Pagefile
         uses: al-cheb/configure-pagefile-action@v1.5
         with:
--- a/.github/workflows/build_pr.yml
+++ b/.github/workflows/build_pr.yml
@@ -16,7 +16,7 @@ jobs:
         with:
           msystem: MINGW64 # D2TM is built within a MinGW64 bit environment
           update: true
-          install: mingw-w64-x86_64-toolchain mingw-w64-x86_64-gcc mingw64/mingw-w64-x86_64-SDL2 mingw64/mingw-w64-x86_64-SDL2_mixer mingw64/mingw-w64-x86_64-SDL2_image mingw64/mingw-w64-x86_64-SDL2_ttf mingw64/mingw-w64-x86_64-cmake mingw-w64-x86_64-catch
+          install: mingw-w64-x86_64-toolchain mingw-w64-x86_64-gcc mingw-w64-x86_64-sdl3 mingw-w64-x86_64-sdl3-mixer mingw-w64-x86_64-sdl3-image mingw-w64-x86_64-sdl3-ttf mingw64/mingw-w64-x86_64-cmake mingw-w64-x86_64-catch
       - name: configure Pagefile
         uses: al-cheb/configure-pagefile-action@v1.5
         with:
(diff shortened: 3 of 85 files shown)


stefanhendriks:
## Why the explicit `<cmath>` includes were added

`d2tm_math.cpp` and `cSelectYourNextConquestState.cpp` use math functions (`std::atan2`, `std::fabs`, `std::sqrt`, `round`) but never explicitly include `<cmath>`.

On SDL2 this worked because `<SDL2/SDL.h>` on Linux and Windows pulled in system headers (X11, etc.) that transitively included `<cmath>`. This is an accidental dependency — the code was relying on a side effect of the SDL2 headers rather than declaring its own dependency.

SDL3's headers are leaner and don't drag in those same system headers, so the transitive `<cmath>` path disappears. The fix is to make the dependency explicit: add `#include <cmath>` directly in the files that need it. This is the correct thing to do regardless of SDL version.

stefanhendriks:
## Fix for the Linux segfault (`createTextureFromIndexedSurfaceWithPalette`)

**Crash:** `SIGSEGV` at line 370 — `palette->ncolors` where `palette` is `nullptr`.

**Root cause:** The old code called `SDL_GetSurfacePalette(modifiableSurface)` to get the palette of the freshly created surface, then dereferenced the result without checking for null. On macOS/Windows this worked because SDL3 auto-creates a default palette when you call `SDL_CreateSurface` with an indexed pixel format. On Linux (at least with some SDL3 builds) that auto-creation doesn't reliably happen, so `SDL_GetSurfacePalette` returns `nullptr`.

**Fix:** Don't rely on getting back the auto-created palette. Instead:
1. Call `SDL_CreatePalette(n)` to explicitly create a palette.
2. Copy the reference surface's colors into it with `SDL_SetPaletteColors`.
3. Apply the house colour swap in that same palette object.
4. Attach it to `modifiableSurface` with `SDL_SetSurfacePalette`.
5. Release our reference with `SDL_DestroyPalette` — the surface now holds its own ref.

This is also cleaner: the entire palette setup happens in one place with one owned pointer, instead of calling `SDL_GetSurfacePalette` three times and hoping the result is always non-null.

stefanhendriks:
## Fix for the shutdown segfault (`SDLTextureDeleter`)

**Crash:** `SIGSEGV` inside `SDL_DestroyTexture` during process teardown.

**Root cause — shutdown ordering:**
The Graphics objects (`gfxdata`, `gfxinter`, etc.) are global `std::shared_ptr`s defined in `main.cpp`. They are destroyed at process exit, which happens **after** `cPlatformLayerInit::~cPlatformLayerInit()` has already called `SDL_Quit()`.

`SDL_Quit()` tears down all SDL subsystems including the renderer. The texture handles inside the Graphics caches are now dangling. When the global destructors run, `SDLTextureDeleter::operator()` calls `SDL_DestroyTexture` on those handles — SDL3 crashes on this, whereas SDL2 was more lenient.

**Fix:** Add a `SDL_WasInit(SDL_INIT_VIDEO)` guard in `SDLTextureDeleter`. If the video subsystem is already gone, skip the destroy call — the OS reclaims the memory on exit anyway. This is the minimal fix that doesn't require restructuring object lifetimes across the whole codebase.

The proper long-term fix would be to ensure all Graphics caches are explicitly cleared before the renderer is destroyed (rather than relying on global destructor order), but that requires larger restructuring and is out of scope for this migration PR.

PR #1251, opened 2026-06-04
cPlatformLayerInit and cScreenInit merged into one class

### LINKED ISSUE #1247: Fusion of 2 class
Can you make fusion about `cPlatformLayerInit` and `cScreenInit` class ?

This made sense before we moved to SDL2 (and even then, it wasn't always the case), and I don't see any reason for it now.

stefanhendriks:
Fixes #1247.

## What changed

`cPlatformLayerInit` was a thin RAII wrapper for `SDL_Init`/`MIX_Init`/`TTF_Init`. It existed as a separate class from `cScreenInit` for historical reasons (the split predates SDL2), but both were always created in sequence and neither made sense without the other.

SDL initialization is now the first step in `cScreenInit`'s constructor, and the SDL shutdown sequence lives in its destructor. `cPlatformLayerInit` is deleted.

## Files changed

- `cScreenInit.h` — removed `#include "cPlatformLayerInit.h"`, made destructor explicit
- `cScreenInit.cpp` — added SDL/mixer/ttf init at the start of the constructor; added destructor with full shutdown
- `cGame.h` — removed `cPlatformLayerInit` forward declaration and `m_PLInit` member
- `cGame_logic.cpp` — removed `#include "game/cPlatformLayerInit.h"` and `m_PLInit` construction; updated stale comment
- `cPlatformLayerInit.h/.cpp` — deleted

## Test plan

- Build cleanly (all 43 tests pass)
- Launch game: SDL init/shutdown log lines appear as before
- Fullscreen toggle works
- Audio plays

Builds cleanly, all 43 unit tests pass.


Mira-Chronos:
This is good work. However, I would change the name of the `cScreenInit` class, which should, for example, be called `cSDLFacade` to accurately reflect its new identity.

stefanhendriks:
@Mira-Chronos naming a class to a pattern is usually a bad idea, can we name it differently and still be clear enough? Maybe Claude can think of a name. ScreenInit is not covering the gist of it right now.

stefanhendriks:
Claude's suggestion: since the class now owns the full SDL subsystem lifecycle (init + window + audio + TTF + shutdown), a few names that describe what it *is* without naming the pattern:

- **cSDLRuntime** — it represents the SDL runtime that boots up and tears down
- **cSDLContext** — it holds everything SDL: window, renderer, and initialized subsystems
- **cSDLSession** — like an open session with SDL from init to quit

My preference is **cSDLContext** — it matches how the rest of the codebase names containers of related state (GameContext, GraphicsContext), and it makes clear that destroying the object ends the SDL session.

stefanhendriks:
Follow-up after looking at how Context is used in the codebase: cSDLContext would actually clash with the existing convention. Here, Context means a dependency bag (cGameObjectContext, cGameControlsContext) — objects that group and provide access to related state. cScreenInit's job is different: it's a RAII lifetime manager — SDL comes up in the constructor, everything tears down in the destructor.

So the better name is **cSDLSystem**. It doesn't name the pattern, it doesn't bend the Context convention, and it accurately describes what it is: the SDL system running in the process, from init to shutdown.

Issue #1254, opened 2026-06-04
Remove global variable game


Mira-Chronos:
The software still relies heavily on references to "game."

game is also a global variable and this is very bad.

The `game` object of type `Game` is defined in the `main.cpp` file and is instantiated very quickly. It is then defined in `dtmc.h` as an external variable and is used throughout the code.

The goal is to replace it with more secure and acceptable access methods using Context classes.

The `GameInterface` class, for example, allows access to certain functions of the `cGame` class.

The `sGameServices` structure allows access to the various services that provide access to the correct object.

There are two ways to do this:

- either pass `sGameService` directly into the class via its constructor.

- or pass `sGameService`, as with the `cMap.h` class, via a function:

`void serviceInit(sGameServices* services);`

stefanhendriks:
## Implementation plan

The `cGame game` instance is declared `extern` in `src/include/d2tmc.h` and accessed directly throughout ~50 files (~842 call sites). The goal is to replace all direct `game.xxx` access with the existing service-injection infrastructure:

- `sGameServices` struct (`src/include/sGameServices.h`) — already adopted by `cUnit`, `cUnits`, `cMap`, `cPlayer`, `cPlayers`, `cGameObjectContext`, `cPathFinder` via `serviceInit(sGameServices*)`
- `cGameInterface` (`src/game/cGameInterface.h`) — facade over `cGame` for UI/sound/camera/events (~30 methods)
- `GameContext` (`src/context/GameContext.hpp`) — infra services (graphics, text, SDL, sound, time)

The most accessed members via `game.`:
- `m_gameObjectsContext` (~449 times) — maps to `sGameServices.objects`
- `m_infoContext` (~106 times) — maps to `sGameServices.info`
- `m_gameSettings` (~81 times) — maps to `sGameServices.settings`
- `m_mapCamera` (~80 times) — not yet in sGameServices; add it when first needed

Steps are ordered by number of `game.` references (simplest first). Each step is one PR. **Extend `sGameServices` only as needed within each step** — do not add fields until a migration step actually requires them.

Reference counts per area (measured):

| Area | Count |
|---|---|
| `src/gamestates/` | 1 |
| `src/gameobjects/cFlag.cpp` | 5 |
| `src/gameobjects/mentat/` | 15 |
| `src/context/` | 17 |
| `src/gameobjects/particles/` | 19 |
| `src/building/` | 24 |
| `src/gameobjects/map/` | 27 |
[...]

PR #1264, opened 2026-06-04
Step 4 of 15: cParticle has zero game. references

stefanhendriks:
## Summary

Step 4 of #1254 — migrates `src/gameobjects/particles/` off the `game` global.

**Infrastructure:**
- `cMapCamera*` added to `sGameServices` (first use) and populated in `cGame_logic` right after camera creation
- `serviceInit(sGameServices*)` added to `cParticle` and `cParticles`; `cGameObjectContext::serviceInit` now calls `m_particles->serviceInit(services)` so all 600 particle slots receive the services pointer at startup

**`cParticle` instance methods** (`draw_x`, `draw_y`, `think_position`, `draw`, `getParticleInfo`, `bindToUnit`, `addPosX/Y`, `die`, `recolorForHouseIfGiven`) now use `m_services->mapCamera`, `->objects`, `->info` instead of the `game` global.

**Static methods** `create()` and `findNewSlot()` now take `cGameObjectContext*` and `cInfoContext*` directly. All call sites updated:
- `cUnit.cpp` uses `m_objects` / `m_infos` (already injected via its own `serviceInit`)
- `cAbstractStructure.cpp`, `cGunTurret.cpp`, `bullet.cpp`, `cMiniMapDrawer.cpp`, `cMouseUnitsSelectedState.cpp` temporarily pass `game.m_gameObjectsContext.get()` / `game.m_infoContext.get()` until their respective migration steps

## Test plan

- [x] `./rebuildmacos.sh` passes (43/43 tests)
- [x] `src/gameobjects/particles/` contains zero `game.` references

### DIFF
--- a/src/gameobjects/particles/cParticle.cpp
+++ b/src/gameobjects/particles/cParticle.cpp
@@ -12,8 +12,8 @@
 #include "cParticle.h"
 #include "utils/RNG.hpp"
 #include "utils/texture_utils.h"
-#include "game/cGame.h"
 #include "include/d2tmc.h"
+#include "include/sGameServices.h"
 #include "drawers/SDLDrawer.hpp"
 #include "gameobjects/map/cMapCamera.h"
 #include "gameobjects/players/cPlayer.h"
@@ -41,6 +41,11 @@ cParticle::~cParticle()
 {
     bmp = nullptr;
     dimensions.reset();
+}
+
+void cParticle::serviceInit(sGameServices* services)
+{
+    m_services = services;
 }
 
 void cParticle::reset()
@@ -85,11 +90,11 @@ bool cParticle::isValid() const {
 }
 
 int cParticle::draw_x() const {
-    return game.m_mapCamera->getWindowXPositionWithOffset(x, drawXBmpOffset);
+    return m_services->mapCamera->getWindowXPositionWithOffset(x, drawXBmpOffset);
 }
 
 int cParticle::draw_y() const {
-    return game.m_mapCamera->getWindowYPositionWithOffset(y, drawYBmpOffset);
+    return m_services->mapCamera->getWindowYPositionWithOffset(y, drawYBmpOffset);
 }
 
 /**
@@ -99,7 +104,7 @@ int cParticle::draw_y() const {
 void cParticle::think_position()
 {
     if (boundUnitID > -1) {
-        cUnit *pUnit = game.m_gameObjectsContext->getUnit(boundUnitID);
+        cUnit *pUnit = m_services->objects->getUnit(boundUnitID);
         if (!pUnit->isValid()) {
             bindToUnit(-1);
         }
@@ -107,9 +112,9 @@ void cParticle::think_position()
 
     // keep updating dimensions
     dimensions->move(draw_x(), draw_y());
-    if (game.m_mapCamera) {
-        dimensions->resize(game.m_mapCamera->factorZoomLevel(getFrameWidth()),
-                           game.m_mapCamera->factorZoomLevel(getFrameHeight()));
+    if (m_services->mapCamera) {
+        dimensions->resize(m_services->mapCamera->factorZoomLevel(getFrameWidth()),
+                           m_services->mapCamera->factorZoomLevel(getFrameHeight()));
     }
 }
 
@@ -126,8 +131,8 @@ void cParticle::draw()
     int frameHeight = getFrameHeight();
 
     // create proper sized bitmap
-    int bmp_width = game.m_mapCamera->factorZoomLevel(frameWidth);
-    int bmp_height = game.m_mapCamera->factorZoomLevel(frameHeight);
+    int bmp_width = m_services->mapCamera->factorZoomLevel(frameWidth);
+    int bmp_height = m_services->mapCamera->factorZoomLevel(frameHeight);
 
     int drawX = draw_x();
     int drawY = draw_y();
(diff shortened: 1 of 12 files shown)


src/gameobjects/particles/cParticle.cpp
```diff
@@ -43,6 +43,11 @@ cParticle::~cParticle()
     dimensions.reset();
 }
 
+void cParticle::serviceInit(sGameServices* services)
```
stefanhendriks, line 47: why is services not passed in via constructor?

src/gameobjects/particles/cParticle.cpp
```diff
@@ -718,17 +724,17 @@ int cParticle::create(long x, long y, int iType, int iHouse, int iFrame, int iUn
...
     pParticle.recolorForHouseIfGiven();
     return iNewId;
 }
 
-int cParticle::findNewSlot()
+int cParticle::findNewSlot(cGameObjectContext* objects)
 {
     int i = 0;
-    for (auto &particle : game.m_gameObjectsContext->getParticles()) {
+    for (auto &particle : objects->getParticles()) {
```
stefanhendriks: this seems to only need the `cParticles` class, can't we pass this in directly instead? Also, it feels like findNewSlot (the whole management of adding particles in a collection or not), should be part of `cParticles`. @claude create ticket for future

src/gameobjects/particles/cParticle.cpp
```diff
@@ -795,7 +801,7 @@ void cParticle::addPosX(float d)
 {
     this->x += d;
     if (boundParticleID > -1) {
-        cParticle &otherParticle = game.m_gameObjectsContext->getParticles()[boundParticleID];
+        cParticle &otherParticle = m_services->objects->getParticles()[boundParticleID];
```
stefanhendriks: can't we extract this logic in a method to fetch particle by id? See also below.

src/gameobjects/particles/cParticle.cpp
```diff
@@ -809,7 +815,7 @@ void cParticle::addPosY(float d)
 {
     this->y += d;
     if (boundParticleID > -1) {
-        cParticle &otherParticle = game.m_gameObjectsContext->getParticles()[boundParticleID];
+        cParticle &otherParticle = m_services->objects->getParticles()[boundParticleID];
```
stefanhendriks: can be extracted

src/gameobjects/particles/cParticle.cpp
```diff
@@ -824,7 +830,7 @@ void cParticle::die()
     bindToUnit(-1);
     bAlive = false;
     if (boundParticleID > -1) {
-        cParticle &pParticle = game.m_gameObjectsContext->getParticles()[boundParticleID];
+        cParticle &pParticle = m_services->objects->getParticles()[boundParticleID];
```
stefanhendriks: can be extracted

src/gameobjects/particles/cParticles.h
```diff
@@ -36,6 +38,12 @@ class cParticles {
         cParticle::reset();
     }
 
+    void serviceInit(sGameServices* services) {
+        for (auto &particle : m_particles) {
+            particle.serviceInit(services);
```
stefanhendriks, line 43: We need to be careful, as newly created particles might not have this service? Make sure it is set properly.

src/gameobjects/particles/cParticle.cpp
stefanhendriks: All 600 `cParticle` instances live in a `std::array` inside `cParticles`, so they're default-constructed before services exist. Using a two-phase `serviceInit` is the established pattern in this codebase (`cUnit`, `cMap`, `cPlayers` all do the same). Passing services via constructor would require either not using a fixed array or changing the array initialisation order.

src/gameobjects/particles/cParticle.cpp
stefanhendriks: Fixed: `findNewSlot` now takes `cParticles&` directly (called with `objects->getParticles()` inside `create()`). Created issue #1265 to track moving `findNewSlot` (and eventually particle creation) into `cParticles`.

src/gameobjects/particles/cParticle.cpp
stefanhendriks: Fixed: extracted `getBoundParticle()` private helper. All three call sites (`addPosX`, `addPosY`, `die`) now use it.

src/gameobjects/particles/cParticle.cpp
stefanhendriks: Fixed: extracted `getBoundParticle()` private helper. All three call sites (`addPosX`, `addPosY`, `die`) now use it.

src/gameobjects/particles/cParticle.cpp
stefanhendriks: Fixed: extracted `getBoundParticle()` private helper. All three call sites (`addPosX`, `addPosY`, `die`) now use it.

src/gameobjects/particles/cParticles.h
stefanhendriks: Safe: all 600 slots are pre-allocated in the `std::array` and `cParticles::serviceInit` iterates every one of them, so every slot has `m_services` set before any `create()` call can happen. `cParticle::init()` (called from `create()`) only resets game state (position, timers, bAlive) — it does not touch `m_services`.

src/gameobjects/particles/cParticle.cpp
```diff
@@ -801,7 +807,7 @@ void cParticle::addPosX(float d)
 {
     this->x += d;
     if (boundParticleID > -1) {
-        cParticle &otherParticle = m_services->objects->getParticles()[boundParticleID];
+        cParticle &otherParticle = getBoundParticle();
```
stefanhendriks: nit:

```cpp
cParticle &boundParticle = getBoundParticle();
```

Same goes for other references

stefanhendriks: Fixed — all three local variables are now `boundParticle`.

PR #1268, opened 2026-06-04
Step 6 of 15: cMapCamera and cRandomMapGenerator have zero game. references

stefanhendriks:
## Summary

- `cMapCamera` constructor gains `cGameSettings*` and `cMouse*` parameters; both stored as members (`m_gameSettings`, `m_mouse`)
- All `game.getMouse()` calls replaced with `m_mouse`
- All `game.m_gameSettings->` calls replaced with `m_gameSettings->`
- `game.m_gameObjectsContext->getMap()->getWidth()/getHeight()/getMaxCells()` replaced with `m_pMap->` (already a member)
- `game.m_mapCamera->zoomIn()/zoomOut()` replaced with `zoomIn()/zoomOut()` (the camera was calling itself through the global)
- `cRandomMapGenerator::generateRandomMap` gains a `cGameObjectContext* objects` parameter; all `game.m_gameObjectsContext->` replaced with `objects->`
- Call site in `cSetupSkirmishState` passes `m_objects` (already available)
- `cGame_logic.cpp` passes `m_gameSettings.get(), getMouse()` to updated `cMapCamera` constructor

Part of #1254 (step 6).

### DIFF
--- a/src/game/cGame_logic.cpp
+++ b/src/game/cGame_logic.cpp
@@ -747,7 +747,7 @@ bool cGame::setupGame()
     if (m_mapCamera != nullptr)
         delete m_mapCamera;
 
-    m_mapCamera = new cMapCamera(m_gameObjectsContext->getMap(), m_cameraDragMoveSpeed, m_cameraBorderOrKeyMoveSpeed, m_cameraEdgeMove);
+    m_mapCamera = new cMapCamera(m_gameObjectsContext->getMap(), m_cameraDragMoveSpeed, m_cameraBorderOrKeyMoveSpeed, m_cameraEdgeMove, m_gameSettings.get(), getMouse());
     m_services->mapCamera = m_mapCamera;
 
     m_cIni->installGame(m_gameFilename);
--- a/src/game/cGame_logic.cpp
+++ b/src/game/cGame_logic.cpp
@@ -747,7 +747,7 @@ bool cGame::setupGame()
     if (m_mapCamera != nullptr)
         delete m_mapCamera;
 
-    m_mapCamera = new cMapCamera(m_gameObjectsContext->getMap(), m_cameraDragMoveSpeed, m_cameraBorderOrKeyMoveSpeed, m_cameraEdgeMove, m_gameSettings.get(), getMouse());
+    m_mapCamera = new cMapCamera(m_gameObjectsContext->getMap(), m_cameraDragMoveSpeed, m_cameraBorderOrKeyMoveSpeed, m_cameraEdgeMove, m_gameSettings.get(), m_mouse);
     m_services->mapCamera = m_mapCamera;
 
     m_cIni->installGame(m_gameFilename);
--- a/src/gameobjects/map/cMapCamera.cpp
+++ b/src/gameobjects/map/cMapCamera.cpp
@@ -1,14 +1,11 @@
 #include "cMapCamera.h"
 
+#include "controls/cMouse.h"
 #include "controls/eKeyAction.h"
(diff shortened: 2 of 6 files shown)


src/game/cGame_logic.cpp
```diff
@@ -747,7 +747,7 @@ bool cGame::setupGame()
     if (m_mapCamera != nullptr)
         delete m_mapCamera;
 
-    m_mapCamera = new cMapCamera(m_gameObjectsContext->getMap(), m_cameraDragMoveSpeed, m_cameraBorderOrKeyMoveSpeed, m_cameraEdgeMove);
+    m_mapCamera = new cMapCamera(m_gameObjectsContext->getMap(), m_cameraDragMoveSpeed, m_cameraBorderOrKeyMoveSpeed, m_cameraEdgeMove, m_gameSettings.get(), getMouse());
```
stefanhendriks: `getMouse()` returns `m_mouse`, why not refer directly to `m_mouse` ?

And make sure it is initialized

src/gameobjects/map/cMapCamera.cpp
```diff
@@ -17,23 +14,27 @@ namespace {
...
     m_moveSpeedBorderOrKeys(moveSpeedBorderOrKeys),
     m_cameraEdgeMove(cameraEdgeMove),
     m_pMap(theMap),
-    m_mapGeometry(theMap ? theMap->getGeometry() : nullptr)
+    m_mapGeometry(theMap ? theMap->getGeometry() : nullptr),
+    m_gameSettings(gameSettings),
+    m_mouse(mouse)
 {
     d2tm_assert(theMap != nullptr);
     d2tm_assert(m_mapGeometry != nullptr);
```
stefanhendriks: this assertion does not make sense. Above `m_mapGeometry` can be `nullPtr` if `theMap` is `null`. But, then the assertion would fail? Instead do `theMap->getGeometry()` directly? 

`theMap` is also asserted to be `!= nullptr`, so this does not make sense whatsoever. Simplify.

src/gameobjects/map/cMapCamera.cpp
```diff
@@ -221,10 +220,10 @@ void cMapCamera::onNotifyMouseEvent(const s_MouseEvent &event)
             onMouseMovedTo(event);
             break;
         case eMouseEventType::MOUSE_SCROLLED_DOWN:
-            game.m_mapCamera->zoomOut();
+            zoomOut();
```
stefanhendriks, line 222: nice

src/gameobjects/map/cMapCamera.cpp
```diff
@@ -296,7 +293,7 @@ void cMapCamera::onMouseRightButtonPressed(const s_MouseEvent &)
     m_moveX = 0.0f;
     m_moveY = 0.0f;
 
-    cMouse *pMouse = game.getMouse();
+    cMouse *pMouse = m_mouse;
```
stefanhendriks: why not use `m_mouse` directly?

src/gameobjects/map/cMapCamera.cpp
```diff
@@ -344,7 +341,7 @@ void cMapCamera::setMoveY(float y, float moveSpeed)
 
 void cMapCamera::onKeyHold(const cKeyboardEvent &event)
 {
-    cMouse *pMouse = game.getMouse();
+    cMouse *pMouse = m_mouse;
```
stefanhendriks: why not use `m_mouse` directly? (is it used?)

PR #1272, opened 2026-06-05
Log improvement (Work in progress)

Mira-Chronos:
## **Goal**

remove old `cLog` class how didn't make dev happy

### Changes
- `cLog` use hard coded singleton
- `Log` is better
- easy write to easy log

```c++
cLogger::getInstance()->log(LOG_ERROR, COMP_SETUP, "CreatorState",
                                     std::format("Forget gameState {}",gameStateToString(gameState)));
```

become:
```c++
Logger::error(COMP_SETUP, "CreatorState", "Forget gameState {}",gameStateToString(gameState));
```


Mira-Chronos:
### next step:
Goal we need to rip `void logbook(const std::string &txt);` from file `utils/common.h`

**Why ?**
This function breaks log analysis and does not allow grouping actions because everything is `WARN` with it.

**How ?**

By reading the logbook comment, you can deduce which `eLogLevel` it belongs to. If you don't know, you can place it in INFO.

By reading the logbook comment, you can deduce which `eLogComponent` it belongs to. You can create your own `eLogComponent` if needed; otherwise, if in doubt, use the `COMP_NONE` component. We can work on it together later.

for `event` string, indicate function name prefixed with his Class if you didn't see what to write

Do not repeat yourself with event and message. Split it into two parts if necessery.

Rip all std::format because, it's integrated to Logger

**Exemple:**
```C++
        logbook(std::format("cMouse::setTile(): Changing mouse tile from {} ({}) to {} ({})", m_mouseTile,
                            mouseTileName(m_mouseTile).c_str(), value, mouseTileName(value).c_str()));
```

become

```C++
        Logger::info(COMP_NONE, "cMouse::setTile", "Changing mouse tile from {} ({}) to {} ({})", m_mouseTile,
                            mouseTileName(m_mouseTile), value, mouseTileName(value) );
```

Issue #1294, opened 2026-06-06
Analyze and document dependency instantiation order


stefanhendriks:
## Problem

The current initialization sequence is implicit and fragile. The `cTextDrawer` crash (startup crash introduced during the game global migration) exposed a real risk: objects that depend on services injected later (via `serviceInit` or setters) can be used before those services are set.

The `cTextDrawer` constructor now requires `cGameSettings*`, but this was only discovered at runtime. Other classes may have similar latent issues.

## Goal

Map the full instantiation order of the major subsystems and identify:

- Which objects are created early (fonts, graphics, settings, context)
- Which objects depend on others that may not yet exist at their construction time
- Where two-phase init (`serviceInit`) is used and whether the ordering guarantee holds
- Any circular or fragile dependencies

## Scope

Focus on `cGame_logic.cpp` init sequence, `GameContext`, `ContextCreator`, `sGameServices`, and all `serviceInit` call sites.

## Expected output

A clear picture (could be a doc, a comment, or just conclusions in this ticket) of what depends on what, and whether any other classes are at risk of the same null-pointer crash pattern.

PR #1296, opened 2026-06-06
Step 15: migrate remaining game global references

stefanhendriks:
## Summary

Final migration step for issue #1254. Removes all remaining `game.` global references outside of `main.cpp`, `cGame_logic.cpp`, and `cGameInterface.cpp` (the three files that legitimately own or wrap `cGame`).

- `cGame::initSkirmish()` was calling `game.missionInit()` through the global inside its own method, and was incorrectly marked `const` — fixed both
- `cGameConditionChecker` stored `m_game` but used the global instead; swapped `cGame*` for `cGameObjectContext*`
- `cTimeManager` swapped `cGame*` for `cGameInterface*`; four `think*()` wrappers added to `cGameInterface`
- `ini.cpp` — `game.missionInit()`, `game.setWinFlags/setLoseFlags`, `game.m_mapCamera`, `game.m_drawManager` all replaced via `m_interface`; `cGame.h` and `d2tmc.h` removed
- `cUnit.cpp`, `texture_utils.cpp`, `GuiButton.cpp` — `cGame.h` removed (was unused after prior steps)
- Dead commented-out code in `cUnit.cpp` referencing `game.map_width`/`game.map_height` removed

Closes part of #1254.

## Test plan

- [ ] Game starts without crash
- [ ] Main menu renders correctly
- [ ] Campaign mission loads and plays correctly (win/lose conditions work)
- [ ] Skirmish game initialises and plays correctly

### DIFF
--- a/src/game/cGameConditionChecker.h
+++ b/src/game/cGameConditionChecker.h
@@ -2,12 +2,12 @@
 
 #include <cstdint>
 
-class cGame;
+class cGameObjectContext;
 class cPlayer;
 
 class cGameConditionChecker {
 public:
-    cGameConditionChecker(cGame* game);
+    cGameConditionChecker(cGameObjectContext* objects);
     bool isMissionWon() const;
     bool isMissionFailed() const;
     void missionInit();
@@ -27,7 +27,7 @@ private:
 
     [[nodiscard]] bool hasGameOverConditionAIHasNoBuildings() const;
     
-    cGame* m_game;
+    cGameObjectContext* m_objects;
     // win/lose flags
     int8_t m_winFlags, m_loseFlags;
 };
--- a/src/game/cGameInterface.cpp
+++ b/src/game/cGameInterface.cpp
@@ -284,3 +284,23 @@ cGameSettings* cGameInterface::getGameSettings() const
 {
     return m_game->m_gameSettings.get();
 }
+
+void cGameInterface::thinkFast() const
+{
+    m_game->thinkFast();
+}
+
+void cGameInterface::thinkNormal() const
+{
+    m_game->thinkNormal();
+}
+
+void cGameInterface::thinkSlow() const
+{
+    m_game->thinkSlow();
+}
+
+void cGameInterface::thinkCache() const
+{
+    m_game->thinkCache();
+}
--- a/src/game/cGameInterface.cpp
+++ b/src/game/cGameInterface.cpp
@@ -284,23 +284,3 @@ cGameSettings* cGameInterface::getGameSettings() const
 {
     return m_game->m_gameSettings.get();
 }
-
-void cGameInterface::thinkFast() const
-{
-    m_game->thinkFast();
-}
-
-void cGameInterface::thinkNormal() const
-{
-    m_game->thinkNormal();
-}
-
-void cGameInterface::thinkSlow() const
-{
-    m_game->thinkSlow();
-}
-
-void cGameInterface::thinkCache() const
-{
-    m_game->thinkCache();
-}
--- a/src/game/cGameInterface.h
+++ b/src/game/cGameInterface.h
@@ -36,6 +36,11 @@ public:
     void missionInit() const;
     void setupPlayers() const;
 
+    void thinkFast() const;
+    void thinkNormal() const;
+    void thinkSlow() const;
+    void thinkCache() const;
+
     void emitGameEvent(const s_GameEvent &event) const;
     void onNotifyGameEvent(const s_GameEvent &event) const;
     void addNotification(const std::string &message, eNotificationType type) const;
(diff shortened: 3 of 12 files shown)


src/game/cGameInterface.h
```diff
@@ -36,6 +36,11 @@ class cGameInterface
     void missionInit() const;
     void setupPlayers() const;
 
+    void thinkFast() const;
```
Mira-Chronos: This is a temporary class; this function is not useful.

src/game/cGameInterface.h
```diff
@@ -36,6 +36,11 @@ class cGameInterface
...
+    void thinkNormal() const;
```
Mira-Chronos: This is a temporary class; this function is not useful.

src/game/cGameInterface.h
```diff
@@ -36,6 +36,11 @@ class cGameInterface
...
+    void thinkSlow() const;
```
Mira-Chronos: This is a temporary class; this function is not useful.

src/game/cGameInterface.h
```diff
@@ -36,6 +36,11 @@ class cGameInterface
...
+    void thinkCache() const;
```
Mira-Chronos: This is a temporary class; this function is not useful.

src/game/cTimeManager.h
```diff
@@ -15,16 +15,17 @@
 #include <stdint.h>
 #include <string>
 #include <memory>
+#include <SDL3/SDL.h>
 
-class cGame;
+class cGameInterface;
```
Mira-Chronos: cTimeManager  is an essential component of cGame, there is no need to go through cGameInterface, there is no point in creating an abstraction.

src/game/cGameInterface.cpp
```diff
@@ -283,4 +283,24 @@ cPlayer* cGameInterface::getPlayer(int id) const
 cGameSettings* cGameInterface::getGameSettings() const
 {
     return m_game->m_gameSettings.get();
+}
+
+void cGameInterface::thinkFast() const
```
Mira-Chronos: Useless code that adds nothing

src/game/cGameInterface.cpp
```diff
@@ -283,4 +283,24 @@ cPlayer* cGameInterface::getPlayer(int id) const
...
+{
+    m_game->thinkFast();
+}
+
+void cGameInterface::thinkNormal() const
```
Mira-Chronos: Useless code that adds nothing

src/game/cGameInterface.cpp
```diff
@@ -283,4 +283,24 @@ cPlayer* cGameInterface::getPlayer(int id) const
...
+{
+    m_game->thinkNormal();
+}
+
+void cGameInterface::thinkSlow() const
```
Mira-Chronos: Useless code that adds nothing

src/game/cGameInterface.cpp
```diff
@@ -283,4 +283,24 @@ cPlayer* cGameInterface::getPlayer(int id) const
...
+{
+    m_game->thinkSlow();
+}
+
+void cGameInterface::thinkCache() const
```
Mira-Chronos: Useless code that adds nothing

src/game/cGameConditionChecker.h
```diff
@@ -2,12 +2,12 @@
 
 #include <cstdint>
 
-class cGame;
+class cGameObjectContext;
```
Mira-Chronos, line 5: This class is a component of cGame and does not require the proposed abstraction. cGame (as a parameter of the class) has its proper place.

Mira-Chronos (changes requested):
In my opinion, class `cTimeManager` and class `cGameinterface` need to revert to a previous state. `cTimeManager`  doesn't need `cGameinterface`, and `cGameinterface`, being a temporary state, doesn't need these new functions that simply shift the code elsewhere.

stefanhendriks:
Good call Mira. Reverted both changes. cTimeManager is back to cGame* and the four think*() wrappers are removed from cGameInterface. The coupling there is intentional -- cTimeManager drives the game loop and cGame owns it.

stefanhendriks:
This PR also surfaced a real instance of the problem described in #1294. When cGameConditionChecker was switched from cGame* to cGameObjectContext*, the assert on construction immediately caught that m_gameObjectsContext was not yet initialised at that point in cGame::init(). The old code masked it because passing 'this' (a valid cGame*) always succeeded, even though the objects context wasn't ready. Fixed by moving the construction site to after m_gameObjectsContext is created.

Mira-Chronos (approved):

PR #1320, opened 2026-06-09
Saboteur can enter and destroy enemy buildings

stefanhendriks:
## Summary

- Saboteur can now be ordered onto enemy structures and destroy them on entry
- `attackAt()` guard that blocked Saboteur (`if (!isAttackingUnit())`) is widened to also allow units where `getUnitInfo().attackIsEnterStructure` is true
- The path-fail else-branch no longer clears `iStructureID` when the goal cell holds the target structure — preserving the enter intent through path retries
- All checks that previously used `isSaboteur()` (a unit-type test) now read `getUnitInfo().attackIsEnterStructure` so any future unit type with that flag follows the same path
- `canAttackUnits` remains `false` so the AI does not treat the Saboteur as a combat unit or include it in attack groups

The flag is not yet configurable via `game.ini` — see follow-up issue for that work.

## Test plan

- [ ] Select a Saboteur unit, hover over an enemy structure — cursor should change to the attack cursor
- [ ] Left-click the enemy structure — Saboteur should move to it and destroy it on entry
- [ ] Confirm the AI does not send Saboteurs as part of normal attack waves
- [ ] Confirm Saboteur cannot be ordered to attack enemy units directly

### DIFF
--- a/src/gameobjects/units/cUnit.cpp
+++ b/src/gameobjects/units/cUnit.cpp
@@ -979,7 +979,7 @@ void cUnit::attack(int goalCell, int unitId, int structureId, int attackCell, bo
 
 void cUnit::attackAt(int cell)
 {
-    if (!isAttackingUnit()) {
+    if (!isAttackingUnit() && !isSaboteur()) {
         return;
     }
 
--- a/src/gameobjects/units/cUnit.cpp
+++ b/src/gameobjects/units/cUnit.cpp
@@ -3059,7 +3059,7 @@ eUnitMoveToCellResult cUnit::moveToNextCellLogic()
                 if (potentialDeadUnit->position.iCell != position.iCell) continue; // not on my cell
                 if (!potentialDeadUnit->canBeSquished()) continue;
 
-                if (potentialDeadUnit->canAttackOnEnterStructure()) {
+                if (potentialDeadUnit->isSaboteur()) {
                     takeDamage(potentialDeadUnit->getUnitInfo().damageOnEnterStructure);
                 }
 
--- a/src/gameobjects/units/cUnit.cpp
+++ b/src/gameobjects/units/cUnit.cpp
@@ -2612,17 +2612,22 @@ void cUnit::thinkFast_move()
                             movement.iPathFails = 0;
                         }
                         else {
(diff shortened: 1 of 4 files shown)


Mira-Chronos:
Hum we change the behavior of this unit if he can attack unit. 
@stefanhendriks  did we need another flag to activate this improvement ?

Mira-Chronos (dismissed):
Wait before : Saboteur worded correctly on v0.7 ?

stefanhendriks:
Afaik Saboteur worked on v0.7.0. Not sure about v0.8.0 @Ripps1 ?

Ripps1:
> Afaik Saboteur worked on v0.7.0. Not sure about v0.8.0 @Ripps1 ?

i can remember roughly when it broke, around the same time we prevented AI from attacking their own stuff

src/gameobjects/units/cUnit.cpp
```diff
@@ -975,7 +975,7 @@ void cUnit::attack(int goalCell, int unitId, int structureId, int attackCell, bo
 
 void cUnit::attackAt(int cell)
 {
-    if (!isAttackingUnit()) {
+    if (!isAttackingUnit() && !isSaboteur()) {
```
Mira-Chronos: This is a hack ?

src/gameobjects/units/cUnit.cpp
```diff
@@ -3061,8 +3058,7 @@ eUnitMoveToCellResult cUnit::moveToNextCellLogic()
                 if (potentialDeadUnit->position.iCell != position.iCell) continue; // not on my cell
                 if (!potentialDeadUnit->canBeSquished()) continue;
 
-                if (potentialDeadUnit->isSaboteur()) {
-                    // this unit takes damage, catches the explosion so to speak
+                if (potentialDeadUnit->canAttackOnEnterStructure()) {
                     takeDamage(potentialDeadUnit->getUnitInfo().damageOnEnterStructure);
```
stefanhendriks, line 3068: we intend for the unit to die. Instead we should let it take damage of "current + 1" ensuring it will trigger death.

Ripps1:
Confirming this has fixed the issue

PR #1323, opened 2026-06-09
Briefing music no longer crashes when house is not yet selected

stefanhendriks:
## Root cause

`BeneMentat::onYesButtonPressed()` sets the human player's house to `GENERALHOUSE` and immediately queues a `GAME_BRIEFING` state transition. When that state starts, `playMusicByType(MUSIC_BRIEFING)` reads `humanPlayer->getHouse()` — which is still `GENERALHOUSE` — and hit a `d2tm_assert` because no mentat track was defined for it.

The same crash can occur with `SARDAUKAR` once that house is fully integrated, since it also has no dedicated mentat track yet.

## Changes

- `GENERALHOUSE` now falls back to `MIDI_MENU` (house selection is still in progress at that point)
- `SARDAUKAR` now falls back to `MIDI_MENU` with a note that a dedicated track can be added later
- Any other undefined house retains the `d2tm_assert` so new houses don't silently play wrong music

## Test plan

- [ ] Click Yes on the BeneMentat screen — game should transition to briefing without crash
- [ ] Click No on the BeneMentat screen — game should return to house selection without crash
- [ ] Start a campaign as Atreides, Harkonnen, and Ordos — confirm correct mentat tracks play

### DIFF
--- a/src/game/cGame_logic.cpp
+++ b/src/game/cGame_logic.cpp
@@ -1499,9 +1499,16 @@ bool cGame::playMusicByType(int iType, int playerId, bool triggerWithVoice)
         else if (houseIndex == ORDOS) {
             sampleId = MIDI_MENTAT_ORD;
         }
-        else {
-            logbook(std::format("playMusicByType: MUSIC_BRIEFING requested for unrecognized house {}, falling back to menu music", houseIndex));
+        else if (houseIndex == SARDAUKAR) {
+            // No dedicated Sardaukar mentat track yet; use menu music as placeholder
             sampleId = MIDI_MENU;
+        }
+        else if (houseIndex == GENERALHOUSE) {
+            // BeneMentat has no dedicated track; house selection is still in progress
+            sampleId = MIDI_MENU;
+        }
+        else {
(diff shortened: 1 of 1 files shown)


Mira-Chronos:
We should test this improvement in Game  to check if we are correctly redirected to the bene gesserit

**BUT:** We're changing the gameplay because before the bug, the music wasn't the issue when clicking "No." Personally, I feel like this bug fixes the symptom, not the cause.

I suggest postponing its implementation because the assertion belongs in the tests for this function.

Mira-Chronos (changes requested):
Wait to see when SardaukarHouse is integrated to identify the cause, not the symptom

stefanhendriks:
@claude dig deeper and analyse what could be the root cause of this issue. Clearly the fix is not sufficient.

Mira-Chronos:
I'm against this PR. If we define a new faction, it's obvious the assertion is useless, but showing that we've done some partial work is pointless. We should really just leave the assertion and let the game crash because we're missing files.

After completing the Sardaukar campaign, I ended up with working code.

The assertion is important, the code itself is important as it is, and I don't want to change it.

stefanhendriks:
Agreed

PR #1344, opened 2026-06-10
Improve reinforcement

Mira-Chronos:
For Ripps proposal:

## **Goal**

Improve loading reinforcement 
- ability to send reinforcements to a given cell
- no more distinction between uppercase and lowercase letters


src/utils/ini.cpp
```diff
@@ -947,8 +946,16 @@ void cIni::INI_Scenario_Section_Reinforcements(int iHouse, const std::string& sl
                         // computer player must find enemy = human
                         targetCell = m_objects->getPlayer(0)->getFocusCell();
                     }
+                } else {
+                    auto sChunk = std::string(chunk);
+                    if (!sChunk.empty() &&
```
stefanhendriks, line 951: ok, so basically we're allowing to set a cell and be that the 'target' of the reinforcement instead. Nice, this seems forwards-compatible and not break existing scenarios 👍

stefanhendriks (approved):

PR #1351, opened 2026-06-10
Cheat mode resets to disabled when returning to main menu

### LINKED ISSUE #1103: Cheat code remains enabled when game ends, can pass over into editor.
Cheat code remains enabled when game ends, can pass over into editor state too

Image

stefanhendriks:
## Summary

Cheat mode was never reset between sessions. Once enabled it would persist through game-over, into the editor, and carry over into subsequent games.

Fix: reset `m_cheatMode = false` on transition to `GAME_MENU` and on transition to `GAME_BRIEFING` (mission select path).

## Test plan

- [ ] Enable cheat mode during a skirmish game
- [ ] Return to main menu — confirm cheat mode is off
- [ ] Start a new game — confirm cheat mode is still off
- [ ] Confirm cheat mode does not bleed into the editor
- [ ] Enable cheat mode mid-mission, select next mission from mission select — confirm cheat mode is off on briefing screen

Closes #1103

### DIFF
--- a/src/game/cGame_logic.cpp
+++ b/src/game/cGame_logic.cpp
@@ -1006,6 +1006,7 @@ void cGame::setState(int newState)
                 playMusicByTypeForStateTransition(MUSIC_BRIEFING);
             }
             else if (newState == GAME_BRIEFING) {
+                m_gameSettings->m_cheatMode = false;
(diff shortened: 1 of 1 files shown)


stefanhendriks:
Should also be reset when jumping to mission, via mission select. So reset during briefing might also be needed.

stefanhendriks:
Added reset on  entry as well, which covers the mission-select path.

PR #1352, opened 2026-06-10
Game no longer hangs when all players are assigned to random house in skirmish

### LINKED ISSUE #1058: Game will crash with 4 or more Random Players on skirmish
If you assign 4 or more Random Players on skirmish the Game will crash.

lets say for arguments sake that when the game starts it looks down the list of players and gives each one a house randomly.
it goes from the top to bottom in order assigning each player a house
it does player 1,
player 2,
player 3,
player 4,
runs out of houses and says fuck
i don't know how to handle this .

stefanhendriks:
## Summary

There are only 4 playable houses (Harkonnen, Atreides, Ordos, Sardaukar) but 5 playable player slots. With all 5 set to Random, the 5th player could never find an unused house and would spin in an infinite loop, freezing the game.

Fix: build the set of available houses upfront. If none are left (all 4 taken), log a warning and assign a random house anyway — allowing a duplicate house/colour — rather than skipping the player slot.

## Test plan

- [ ] Start a skirmish with 5 Random players — game should start normally with all 5 active (one will share a house colour)
- [ ] Confirm the warning appears in the log when all houses are taken
- [ ] Confirm normal skirmish works with 2-4 random players

Closes #1058

### DIFF
--- a/src/gamestates/cSetupSkirmishState.cpp
+++ b/src/gamestates/cSetupSkirmishState.cpp
@@ -26,9 +26,10 @@
 #include "include/Texture.hpp"
 #include "gui/GuiButton.h"
 
-#include <format>
 #include <algorithm>
+#include <format>
 #include <utility>
+#include <vector>
 #include "include/cAssert.h"
 
 #include "config.h"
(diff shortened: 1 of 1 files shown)


src/gamestates/cSetupSkirmishState.cpp
```diff
@@ -726,29 +727,21 @@ void cSetupSkirmishState::prepareSkirmishGameToPlayAndTransitionToCombatState(in
...
+                for (int pl = 0; pl < AI_WORM; pl++) {
+                    int usedHouse = (skirmishPlayer[pl].iHouse > 0)
+                        ? skirmishPlayer[pl].iHouse
+                        : m_objects->getPlayer(pl)->getHouse();
+                    available.erase(
+                        std::remove(available.begin(), available.end(), usedHouse),
+                        available.end());
+                }
+                if (available.empty()) {
+                    cPlayer* pSkipPlayer = m_objects->getPlayer(p);
```
stefanhendriks: add log statement, as this was expected to be a playable faction and somehow did not get a house. Log the current player id's + house setup.

stefanhendriks:
This PR will make it impossible to play with 5 players if any are set as Random house; be mindful. 

Alternatively we can set the house of the unfound player to a random house (even though all houses have been shared already), it will mean a duplicate house and thus color, but still allow for a 5th player to play. This is an edge case, as we don't really support it. Atm it is max 4 players supported.

PR #1354, opened 2026-06-10
Carry-all corpse de-spawns after explosion

### LINKED ISSUE #994: Carry-All Corpse does not de-spawn like ornithopter corpse
The  Carry-All Corpse does not de-spawn like ornithopter corpse does, its actually a cool effect, we should make an option to keep this,
but they should ALWAYS de-spawn when crashing on a building.

Image

stefanhendriks:
## Summary

- `D2TM_PARTICLE_EXPLOSION_CARRYALL` was missing from the two blocks in `cParticle.cpp` that handle fade-out and frame-timer initialization for orni/squish explosion sprites
- The carry-all corpse sprite now fades out and de-spawns the same way the orni explosion already does

Closes #994

### DIFF
--- a/src/gameobjects/particles/cParticle.cpp
+++ b/src/gameobjects/particles/cParticle.cpp
@@ -520,7 +520,8 @@ void cParticle::thinkFast()
     }
 
     if (iType == D2TM_PARTICLE_SQUISH01 || iType == D2TM_PARTICLE_SQUISH02 ||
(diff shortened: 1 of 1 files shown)


stefanhendriks:
Ticket mentions something about having a way to influence particle despawn. Make sure that part is captured in a new ticket

stefanhendriks:
Created #1361 to track the particle despawn configurability mentioned in the original ticket.

Issue #1377, opened 2026-06-12
Game crashes before main menu


Danzemon:
OS Win 11

BUILD: d2tm-windows-5365cb7f 

When opening the executable, my laptop thinks for a few seconds then the game crashes completely BEFORE main menu appears.

Attempted to use -debug flag to bring the log. No luck.

BUILD: d2tm-windows-2049e335

When opening the executable, my laptop thinks for a few seconds then the game crashes completely BEFORE main menu appears.

Attempted to use -debug flag to bring the log. No luck.

OS Win 10 Virtual Machine

BUILD: d2tm-windows-5365cb7f 

When opening the executable, my laptop thinks for a few seconds then the game crashes completely BEFORE main menu appears.

Attempted to use -debug flag to bring the log. No luck.

BUILD: d2tm-windows-2049e335

When opening the executable, my laptop thinks for a few seconds then the game crashes completely BEFORE main menu appears.

Attempted to use -debug flag. Game runs. 

All instances of the crash returns the same message as per the attached log file.

[dump.log](https://github.com/user-attachments/files/28876156/dump.log)

Issue #1253, opened 2026-06-04
Renaming convention on repertory


Mira-Chronos:
The program uses Hungarian notation for source files and classes, which is a long-standing and counterproductive standard, especially with the IDEs available.

The task would be to:
- Rename the class without the 'c' prefix.
- Rename the file to the class name.
- Use `CamelCase` notation for the class name and `camelCase` notation for the class's functions and methods.
- Prefix all class variables with 'm_'.
- Distinguish between header files with the '.h' extension for C code and '.hpp' for C++ code.
- If possible, avoid C++ code in an 'hpp' file.

Mira-Chronos:
After all renaming ... why not using C++ modules ?

PR #1381, opened 2026-06-13
Refactor state creation

Mira-Chronos:
For #1377 

## **Goal**

Using `StateCreator` to avoid memory corruption on Windows 
`StateCreator` is now used after long shadow unused code

if @claude  can make a detailled review ...

### DIFF
--- a/src/game/cGame.h
+++ b/src/game/cGame.h
@@ -247,10 +247,11 @@ private:
     void thinkFast_fading();
 
 
-    int m_state;
-    int m_nextState;
-    // the current game state we are running
+    int m_state; // TODO: Replace with enum, and later even remove with pointer to state
+    int m_nextState; // TODO: Replace with enum, and later even remove with pointer to state
+    
     cGameState *m_currentState;
+
     void transitionStateIfRequired();
     void setState(int newState);
 
--- a/src/game/cGame.h
+++ b/src/game/cGame.h
@@ -61,6 +61,7 @@ class cStructureUtils;
 class Texture;
 class cIni;
 class cEventEmitter;
+class CreatorState;
 
 struct s_DataCampaign;
 // struct s_PreviewMap;
@@ -282,4 +283,5 @@ private:
     std::unique_ptr<sGameServices> m_services;
     std::unique_ptr<cIni> m_cIni;
     std::unique_ptr<cEventEmitter> m_eventEmitter;
+    std::unique_ptr<CreatorState> m_creatorState;
 };
--- a/src/game/cGame.h
+++ b/src/game/cGame.h
@@ -61,7 +61,7 @@ class cStructureUtils;
 class Texture;
 class cIni;
 class cEventEmitter;
-class CreatorState;
+class cCreatorState;
 
 struct s_DataCampaign;
 // struct s_PreviewMap;
@@ -279,5 +279,5 @@ private:
     std::unique_ptr<sGameServices> m_services;
     std::unique_ptr<cIni> m_cIni;
     std::unique_ptr<cEventEmitter> m_eventEmitter;
-    std::unique_ptr<CreatorState> m_creatorState;
+    std::unique_ptr<cCreatorState> m_creatorState;
 };
--- a/src/game/cGame.h
+++ b/src/game/cGame.h
@@ -185,7 +185,6 @@ public:
     void applySettings(std::unique_ptr<InitialGameSettings> gs);
     void changeStateFromMentat();
     void loadMapFromEditor(int map);
-    // void loadMapFromEditor(s_PreviewMap *map);
 
     Texture* getScreenTexture() const {
         return screenTexture;
@@ -217,8 +216,6 @@ private:
     std::unique_ptr<cFocusManager> m_focusManager;
 
     cSoundPlayer* m_soundPlayer;
-
-    // std::shared_ptr<cPreviewMaps> m_PreviewMaps;
 
     std::unique_ptr<InitialGameSettings> m_initialGameSettings;
 
--- a/src/game/cGame.h
+++ b/src/game/cGame.h
@@ -251,7 +251,6 @@ private:
     int m_nextState;
     // the current game state we are running
     cGameState *m_currentState;
-    cGameState *m_states[GAME_MAX_STATES];
     void transitionStateIfRequired();
     void setState(int newState);
 
--- a/src/game/cGame.h
+++ b/src/game/cGame.h
@@ -249,7 +249,7 @@ private:
 
     int m_state; // TODO: Replace with enum, and later even remove with pointer to state
     int m_nextState; // TODO: Replace with enum, and later even remove with pointer to state
-    
+
     cGameState *m_currentState;
 
     void transitionStateIfRequired();
--- a/src/game/cGame_logic.cpp
+++ b/src/game/cGame_logic.cpp
@@ -133,8 +133,6 @@ void cGame::syncTextInputState() const
 
 cGame::cGame()
 {
-    std::fill(std::begin(m_states), std::end(m_states), nullptr);
-
     m_nextState = -1;
     m_currentState = nullptr;
 
@@ -506,6 +504,7 @@ void cGame::run()
 
         updateMouseAndKeyboardState();
         m_currentState->update();
+        //Logger::print(LOG_INFO , COMP_NONE, "m_currentState", "m_currentState {}", gameStateToString(m_currentState->getType()));
 
         m_renderDrawer->beginDrawingToTexture(actualRenderer);
         m_renderDrawer->renderClearToColor();
@@ -534,36 +533,8 @@ void cGame::shutdown()
     m_gameObjectsContext->getParticles().reset();
     Logger::info(COMP_NONE, "cGame::shutdown", "=== SHUTDOWN ===");
 
-    for (int i = 0; i < GAME_MAX_STATES; i++) {
-        cGameState *pState = m_states[i];
-        if (pState) {
-            if (pState->getType() != eGameStateType::GAMESTATE_SELECT_YOUR_NEXT_CONQUEST) {
-                if (m_currentState == pState) {
-                    m_currentState = nullptr; // reset
-                }
-                delete pState;
-                m_states[i] = nullptr;
-            }
-            else {
-                if (m_currentState == pState) {
-                    m_currentState = nullptr; // reset
-                }
-                m_states[i] = nullptr;
-            }
-        }
-    }
-
-    if (m_currentState != nullptr) {
-        d2tm_assert(false);
-        if (m_currentState->getType() != eGameStateType::GAMESTATE_SELECT_YOUR_NEXT_CONQUEST) {
-            // destroy game state object, unless we talk about the region select
-            delete m_currentState;
-        }
-        else {
-            m_currentState = nullptr;
-        }
-    }
-
+    m_currentState = nullptr;
+    m_creatorState->destroyAllStates();
 
     delete m_mapViewport;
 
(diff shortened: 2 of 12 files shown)


stefanhendriks:
*Review posted by Claude on behalf of Stefan.*

---

## Review: Fix 1377 StateCreator integration

Good direction overall — replacing the raw `m_states[GAME_MAX_STATES]` array with `CreatorState` (using `std::optional<std::unique_ptr<cGameState>>` internally) eliminates the manual `new`/`delete` ownership that likely caused the Windows crash. The two latent bugs in `createStateFromScratch` (`NEW_MAP_EDITOR` storing into the wrong slot, `WINNING` creating `Outcome::Lose`) are correctly fixed here too.

That said, a few things need attention before merge:

---

### Issues

**1. Committed debug comment (`cGame_logic.cpp`, in `run()`)**

```cpp
//Logger::print(LOG_INFO , COMP_NONE, "m_currentState", "m_currentState {}", gameStateToString(m_currentState->getType()));
```

Noise comment with no explanatory value. Should be removed before merge.

---

**2. Wrong log message in `prepareMentatToTellAboutHouse`**

```cpp
Logger::warn(COMP_GAME, "cGame::prepareMentatToTellAboutHouse", "cMentatState for TELLHOUSE already exists. Not creating a new one.");
```

The state in question is `cTellHouseState`, not `cMentatState`. This looks like a copy-paste error from a nearby call site.

---

**3. `getState()` called purely for its side effect**

```cpp
if (!m_creatorState->hasState(eGameState::TELLHOUSE)) {
    m_creatorState->getState(eGameState::TELLHOUSE);   // result discarded
    playMusicByTypeForStateTransition(MUSIC_BRIEFING);
}
```

Calling a method named `get*` to create a state without using the return value is surprising. A dedicated `createState(eGameState)` method (or calling `getState` and assigning even if unused via `std::ignore`) would make the intent clearer.

---

**4. Unchecked `dynamic_cast` in `setState` for `GAME_REGION` (new-state branch)**

```cpp
auto *pState = dynamic_cast<cSelectYourNextConquestState *>(m_creatorState->getState(eGameState::REGION));
pState->calculateOffset();
```

If `getState` ever returns null or a mismatched type, this crashes. `jumpToSelectYourNextConquestMission` correctly guards the cast with a null check — same guard would be appropriate here.

---

**5. Missing newline at end of `cCreatorState.cpp`**

Minor, but the file ends without a trailing newline. Most editors/compilers warn about this.

---

### Follow-up (not blocking)

`needToRecreateState` is now always filled with `false`, making the name misleading — it no longer expresses "do we need to recreate?" A follow-up rename to something like `m_forceRecreateOnGet` would reduce confusion.

---

Good work on the migration, Mira. The changes in `shutdown()` especially are a nice simplification. Once the points above are addressed this should be good to go.

src/game/cGame.h
```diff
@@ -61,6 +61,7 @@ class cStructureUtils;
 class Texture;
 class cIni;
 class cEventEmitter;
+class CreatorState;
```
stefanhendriks:
```suggestion
class cCreatorState;
```

src/game/cGame_logic.cpp
```diff
@@ -782,15 +753,13 @@ bool cGame::isState(int thisState) const
 
 void cGame::jumpToSelectYourNextConquestMission(int missionNr)
 {
-    cGameState *existingStatePtr = m_states[GAME_REGION];
-    if (existingStatePtr) {
-        delete existingStatePtr;
-        m_states[GAME_REGION] = nullptr;
+    cGameState *state = m_creatorState->getState(eGameState::REGION, true);
```
stefanhendriks: wouldnt the `cCreatorState` be responsible for creation of a new one? (or returning the existing one)

src/game/cGame_logic.cpp
```diff
@@ -818,38 +787,73 @@ void cGame::setState(int newState)
...
+                case GAME_LOSEBRIEF: mapped = eGameState::LOSEBRIEF; return true;
+                case GAME_LOSING: mapped = eGameState::LOSING; return true;
+                case GAME_SETUPSKIRMISH: mapped = eGameState::SETUPSKIRMISH; return true;
+                case GAME_CREDITS: mapped = eGameState::CREDITS; return true;
+                case GAME_MISSIONSELECT: mapped = eGameState::MISSIONSELECT; return true;
+                case GAME_NEW_MAP_EDITOR: mapped = eGameState::NEW_MAP_EDITOR; return true;
+                default:
+                    return false;
+            }
+        };
```
stefanhendriks, line 817: extract into function `shouldDeleteState` ?

src/gamestates/cCreatorState.h
stefanhendriks:
```suggestion
class cCreatorState {
```
for consistency

src/gamestates/cGameState.h
```diff
@@ -30,6 +30,7 @@ enum eGameStateType {
                                             // #define GAME_WINBRIEF    12      // mentat chatter when won the mission
 };
 
+const char *gameStateToString(eGameStateType state);
```
stefanhendriks: why not part of cGameState itself?

src/gamestates/cCreatorState.cpp
```diff
@@ -27,19 +30,10 @@ CreatorState::CreatorState(sGameServices* services, cIni* ini, s_DataCampaign* d
...
-    needToRecreateState[eGameState::MISSIONSELECT] = false;
-    needToRecreateState[eGameState::MENU] = false;
-    needToRecreateState[eGameState::EDITOR] = false;
-    needToRecreateState[eGameState::NEW_MAP_EDITOR] = false;
-    needToRecreateState[eGameState::WINNING] = false;
-    needToRecreateState[eGameState::LOSING] = false;
+    m_interface = m_services->ctx->getGameInterface();
+    d2tm_assert(m_interface != nullptr);
+    // Reuse states by default. Recreate only when explicitly requested via forceRecreate.
+    needToRecreateState.fill(false);
```
stefanhendriks: `needToRecreateState` seems to be no longer used? can be removed?

src/game/cGame_logic.cpp
```diff
@@ -782,15 +753,13 @@ bool cGame::isState(int thisState) const
...
-    cGameState *existingStatePtr = m_states[GAME_REGION];
-    if (existingStatePtr) {
-        delete existingStatePtr;
-        m_states[GAME_REGION] = nullptr;
+    cGameState *state = m_creatorState->getState(eGameState::REGION, true);
+    auto *pState = dynamic_cast<cSelectYourNextConquestState *>(state);
+    if (!pState) {
+        Logger::error(COMP_GAME, "cGame::jumpToSelectYourNextConquestMission", "Failed to create cSelectYourNextConquestState for mission {}. Aborting.", missionNr);
+        return;
     }
```
stefanhendriks:
```suggestion
    auto *state = m_creatorState->getState(eGameState::REGION, true);
    if (!state) {
        Logger::error(COMP_GAME, "cGame::jumpToSelectYourNextConquestMission", "Failed to create cSelectYourNextConquestState for mission {}. Aborting.", missionNr);
        return;
}
```

src/game/cGame_logic.cpp
```diff
@@ -818,38 +787,73 @@ void cGame::setState(int newState)
...
+        cGameState *existingStatePtr = nullptr;
+        eGameState targetCreatorState = eGameState::MENU;
+        if (mapToCreatorState(newState, targetCreatorState)) {
+            if (deleteOldState) {
+                m_creatorState->destroyState(targetCreatorState);
+            }
+            if (m_creatorState->hasState(targetCreatorState)) {
+                existingStatePtr = m_creatorState->getState(targetCreatorState);
+            }
+        }
```
stefanhendriks, line 860: It feels like we should just have a `getOrCreate` state and a `getNewState`, so that it can delete the old one and re-create the new one. Instead of 'deleteOldState` here? This logic of knowing when to create a new state and thus delete it feels to me a responsibility for the `CreatorState` now.

src/gamestates/cCreatorState.cpp
stefanhendriks:
```suggestion
cGameState* CreatorState::getOrCreateState(eGameState gameState, bool forceRecreate)
```

stefanhendriks (changes requested):

stefanhendriks:
*Recap of Stefan's inline review (posted by Claude on his behalf).*

Stefan submitted a **CHANGES_REQUESTED** review with the following inline notes:

---

### Naming

- **`cGame.h` line 64 / `cCreatorState.h` line 16** — The class is forward-declared and defined as `CreatorState`, which breaks the project's `c`-prefix convention. Suggestion: rename to `cCreatorState` throughout.

---

### Design / responsibility

- **`cCreatorState.cpp` line 45** — `getState` is doing creation too, so the name is misleading. Suggestion: rename to `getOrCreateState`.

- **`cGame_logic.cpp` line 856** — The `deleteOldState` flag and the logic around when to destroy and recreate a state still lives in `cGame::setState`. That responsibility belongs in `CreatorState`. Suggestion: expose a `getOrCreateState` (reuse existing) and a `getNewState` (destroy old, create fresh) so `cGame` doesn't need to know about the deletion logic.

- **`cGame_logic.cpp` line 756** — Related: in `jumpToSelectYourNextConquestMission`, should `CreatorState` itself handle whether to create a new state or return the existing one, rather than the caller passing `forceRecreate = true`?

- **`cGameState.h` line 33** — `gameStateToString(eGameStateType)` is declared as a free function. Why not make it a method on `cGameState` itself?

- **`cGame_logic.cpp` line 813** — The `deleteOldState` boolean expression is complex enough to extract into its own function `shouldDeleteState()`.

---

### Dead code / cleanup

- **`cCreatorState.cpp` line 36** — `needToRecreateState` is now always `fill(false)` and doesn't appear to be read anywhere. Looks like it can be removed entirely.

---

### Correctness

- **`cGame_logic.cpp` line 761** — Suggested simplification for `jumpToSelectYourNextConquestMission`: skip the `dynamic_cast` and just use the raw pointer returned by `getState`, doing a simple null check. (The dynamic cast was already flagged in the earlier Claude review as having no guard.)

---

Overall: good structural direction, but the `CreatorState`/`cGame` responsibility boundary needs tightening before merge.

stefanhendriks:
Related issue opened: #1383 — Extract `cGameStateMachine` to make `setState()` testable and refactorable.

The issue builds on the `cCreatorState` work in this PR. It proposes extracting the state machine logic from `cGame::setState()` into a standalone class so it can be unit tested (and safely refactored) without needing the full SDL/game stack. @Mira-Chronos your input is specifically invited there — see the question at the bottom of the issue.

src/game/cGame_logic.cpp
Mira-Chronos: Yes but as migration was difficult, it's `cGame_logic.cpp` how make it. But in the future, it's to `cCreatorState` to be responsible for creation. Step by step ...

src/game/cGame_logic.cpp
Mira-Chronos: No. Should be removed.

src/gamestates/cCreatorState.cpp
Mira-Chronos: Perhaps yet, but `needToRecreateState` should be used in the future. Step by step make the transition slow.

src/game/cGame_logic.cpp
Mira-Chronos: Yes yes yes ...

src/game/cGame.h
Mira-Chronos: Yes for now. See #1253
Code has been written 3/4 months  before today.

src/game/cGame_logic.cpp
Mira-Chronos: i build new code as close as possible from original to avoid difficulty. I propose a refactor after that to better code.

stefanhendriks (approved):

Issue #1160, opened 2026-05-14
Crash when destroying enemy Conyard


Ripps1:
Not sure what caused this, log is far too large to share on GH

uploaded [here](https://faaf.me/bdf61d61) instead

stefanhendriks:
You can upload a zipped log ;-)

stefanhendriks:
I have seen the `log`, unfortunately I can't find any reason why it crashed. Can you tell anything about how it got destroyed? what were the circumstances?

Ripps1:
> I have seen the `log`, unfortunately I can't find any reason why it crashed. Can you tell anything about how it got destroyed? what were the circumstances?

I've discussed this crash (what i believe to be this crash) a few times with @Mira-Chronos, however each time it happened the circumstances seem to change, for example i could not re-create this specific crash, however there is *somewhere* a problem in the code that causes the game to randomly crash with what seem like repeatable steps but when trying to force it to do it again it wont.

For example there is one that plagues me involving alt+tabbing / switching focus
upon return, and pressing the 'back' button on options screen to resume gameplay results in instant crash.
I've been unable to force or replicate this crash too, it just happens randomly.

stefanhendriks:
@claude can you analyse the code and give an idea what could potentially have caused a crash?

stefanhendriks:
The existing log didn't reveal a clear cause and the crash hasn't been reproduced reliably. To make progress here I need a log from a session where this happened with the Conyard still standing when the crash occurred. If you can reproduce it, please note: what house was it, were there any units spawned from the destruction, and was it an AI or human Conyard? That will help narrow down the `UponDestructionSpawnUnit` path vs. something in the structure removal code.

Ripps1:
more testing required then as no other logs currently exist.

stefanhendriks:
Could this perhaps be related to #1404 ? in both cases a structure gets destroyed.

PR #1392, opened 2026-06-15
setHouse unit recoloring skips unconfigured unit types

stefanhendriks:
## Summary

- `setHouse` iterates all `MAX_UNITTYPES` (50) entries to recreate player-colored textures when a player changes house (eg. on defeat)
- The structure loop already has a `configured` guard; the unit loop had none, so unused slots (which default to the quad bitmap) were also recolored
- Added `bool configured = false;` to `s_UnitInfo`, set it to `true` at the end of each unit's initialisation block in `initUnits`, and guarded the loop with `if (!unitInfo.configured) continue;`
- Also renames the loop variable from `unitType` to `unitInfo`

For #1160

## Test plan

- [ ] Destroy an enemy's last structure in skirmish to trigger player defeat — no crash
- [ ] Verify player-colored unit textures still render correctly after the fix

### DIFF
--- a/src/gameobjects/players/cPlayer.cpp
+++ b/src/gameobjects/players/cPlayer.cpp
@@ -306,15 +306,17 @@ void cPlayer::setHouse(int iHouse)
 
         // same goes for units
         for (int i = 0; i < MAX_UNITTYPES; i++) {
-            s_UnitInfo &unitType = m_infos->getUnitInfo(i);
+            s_UnitInfo &unitInfo = m_infos->getUnitInfo(i);
 
-            bmp_unit[i] = createPlayerTextureFromIndexedSurfaceWithPalette(m_renderer, this, unitType.bmp, TransparentColorIndex);
+            if (!unitInfo.configured) continue;
+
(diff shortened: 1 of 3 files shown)


src/gameobjects/players/cPlayer.cpp
```diff
@@ -308,6 +308,8 @@ void cPlayer::setHouse(int iHouse)
         for (int i = 0; i < MAX_UNITTYPES; i++) {
             s_UnitInfo &unitType = m_infos->getUnitInfo(i);
 
+            if (!unitType.bmp) continue;
```
stefanhendriks: I do not agree with this approach. It is undesirable that a unit info has no `bmp`. It does not make sense to have a unit without `bmp`.

So, instead, check for this when creating unitInfo's. 

If we want to not recolor, we need to validate that a `unitInfo` (not `unitType`, nitpick) should be configured. Perhaps we can add a flag that communicates if the information has been configured?

stefanhendriks:
The intent of this PR is to fix #1160 ,  but it is not the full fix as already identified.

PR #1393, opened 2026-06-15
Main menu no longer responds to clicks during screen fade

### LINKED ISSUE #1194: Spam clicking buttons on the Main Menu screen can result in strangness
https://github.com/user-attachments/assets/f2dcfaf0-3c44-4bd2-8278-d2cab20e9f82

stefanhendriks:
## Summary

- Rapid clicks on the main menu during a fade-out could forward events to the incoming state before it was ready, causing double transitions (eg. two game sessions starting)
- Root cause: `onNotifyMouseEvent` forwarded to the current state unconditionally; `drawState` already blocks rendering during fade-out but event delivery continued
- Fix: added `!m_cScreenFader->isFading()` guard in `onNotifyMouseEvent` so the active state receives no mouse events while a fade is in progress

Fixes #1194

## Test plan

- [ ] On the main menu, click a button quickly and repeatedly while the fade plays — only one transition should occur
- [ ] Verify that normal (non-spammed) button clicks still work as expected

### DIFF
--- a/src/game/cGame_logic.cpp
+++ b/src/game/cGame_logic.cpp
@@ -1233,7 +1233,7 @@ void cGame::onNotifyMouseEvent(const s_MouseEvent &event)
     }
 
     // pass through any classes that are interested
-    if (m_currentState) {
+    if (m_currentState && !m_cScreenFader->isFading()) {
         m_currentState->onNotifyMouseEvent(event);
     }
 


Ripps1:
pretty sure this has broken the main menu, please see [Issue #1397](https://github.com/stefanhendriks/Dune-II---The-Maker/issues/1397)

PR #1395, opened 2026-06-15
Button render cycling via right-click restricted to debug mode

### LINKED ISSUE #1036: Feedback: Skirmish Screen Map Selection 'Next' Button breaks
I've no idea why or what causes this, but sometimes (randomly) it happens.

Image

stefanhendriks:
Closes #1036

Right-clicking any `GuiButton` while focused previously cycled through render kinds (`nextRenderKind()`), changing the button's visual appearance. This was a developer tool that leaked into non-debug builds.

`GuiButton` now holds a static `cGameSettings*` registered once at startup via `GuiButton::registerSettings()`. The `nextRenderKind()` call in `onMouseRightButtonClicked` is gated behind `s_settings->isDebugMode()`, so render cycling only works when the game is launched with debug mode enabled.

## Changes

- `GuiButton.h` / `GuiButton.cpp` — static `s_settings` pointer; `registerSettings()` setter; debug-mode guard in `onMouseRightButtonClicked`
- `cGame_logic.cpp` — calls `GuiButton::registerSettings(m_gameSettings.get())` at end of `applySettings()`

### DIFF
--- a/src/gui/GuiButton.h
+++ b/src/gui/GuiButton.h
@@ -9,6 +9,7 @@
 
 class SDLDrawer;
 class cTextDrawer;
+class cGameSettings;
 
 
 class GuiButton : public GuiObject {
@@ -35,6 +36,8 @@ public:
     void setOnLeftMouseButtonClickedAction(std::function<void()> action);
     void setOnRightMouseButtonClickedAction(std::function<void()> action);
 
+    static void registerSettings(cGameSettings* settings);
+
     void setEnabled(bool value);
 
 private:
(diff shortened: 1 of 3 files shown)


src/gui/GuiButton.h
```diff
@@ -52,6 +55,8 @@ class GuiButton : public GuiObject {
     bool m_pressed;
     bool m_enabled;
 
+    static cGameSettings* s_settings;
```
stefanhendriks, line 58: no static reference here. We should just use m_settings, like everywhere else. Pass in via constructor? Why not?

stefanhendriks:
After consideration, this is already fixed. No need to make code changes.

PR #1406, opened 2026-06-17
Sound slider no longer changes music volume

### LINKED ISSUE #1365: Music Volume is effected by Sound Volume Slider.
The music volume seems to respond to both volume sliders while sound effects only effected by their own.

mvanhorn:
Fixes #1365

## Goal

The music volume responded to both the Music slider and the Sound slider, while sound effects only responded to their own slider. This PR makes the Sound slider affect only sound effects and the Music slider affect only music.

In `cSoundPlayer`, music plays on a dedicated `m_musicTrack` via `MIX_SetTrackGain`, but the sound-effects volume was applied with `MIX_SetMixerGain(m_mixer, ...)`. The master mixer gain modulates every sample passing through the mixer, including the music track, so dragging the Sound slider scaled the music too.

The fix keeps the master mixer gain fixed and gives sound effects their own gain path, so the two sliders no longer interfere.

## Changes

- Keep the master mixer gain fixed at `1.0` so it no longer multiplies the music track.
- Route sound effects through a round-robin pool of `MIX_Track`s (32 slots) and apply the Sound slider value with `MIX_SetTrackGain` on those tracks, so SFX volume is independent of music.
- Call `MIX_StopTrack` before reassigning audio to a pooled track, matching the pattern used for the music track and ensuring clean state on first use.
- Call `missionInit()` after `drawManager->reset()` when re-entering the playing state, so the credits display syncs to the actual balance instead of recounting from zero.

## Testing

- Sound slider to 0, Music slider to max: music plays at full volume; sound effects silent.
- Music slider to 0, Sound slider to max: sound effects play; music silent.
- Both sliders mid-range: moving either one leaves the other channel's volume unchanged.
- Fire multiple units at once: overlapping effects still mix.
- Go to options and back during a skirmish: credits display does not recount from zero.
- Select units while credits are rolling: unit acknowledgement sounds play without being smothered.

### DIFF
--- a/src/game/cGame_logic.cpp
+++ b/src/game/cGame_logic.cpp
@@ -900,6 +900,7 @@ void cGame::setState(int newState)
                     m_currentState = existingStatePtr;
 
                     m_drawManager->reset();
+                    m_drawManager->missionInit();
                     // evaluate all players, so we have initial 'alive' values set properly
                     m_players->evaluateStillAliveForAI();
                     m_gameObjectsContext->getParticles().reset();
--- a/src/utils/cSoundPlayer.cpp
+++ b/src/utils/cSoundPlayer.cpp
@@ -55,6 +55,12 @@ cSoundPlayer::cSoundPlayer(const std::string &datafile)
         m_isMusicEnabled = true;
         m_isSoundEnabled = true;
         m_musicTrack = MIX_CreateTrack(m_mixer);
+        for (int i = 0; i < kSfxTrackPoolSize; ++i) {
+            MIX_Track *track = MIX_CreateTrack(m_mixer);
+            if (track) {
+                m_sfxTracks.push_back(track);
+            }
+        }
     }
 
     soundData = std::make_unique<cSoundData>(datafile, m_mixer);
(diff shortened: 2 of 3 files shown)


Ripps1:
Awesome... much love for the assist. I'm sure Stefan will thank you himself for this.

stefanhendriks:
Tested it and worked. Great!

stefanhendriks:
One nitpick: I noticed, when selecting units. I only hear the 'static' noise but not "Reporting". When I reselect, I hear the full sentence. After that it keeps working. So it seems to only happen when I initiate (first?) mission or set sound volume.

stefanhendriks:
Analysing with Claude why i have this issue. It is consistent and once you notice it, it is hard to not notice it ;-)

stefanhendriks:
there is some progress. I managed to fix the initial cut-off of sounds after changing volume (or initial mission start). But now I notice that when credits are earned that sounds seem to be cut-off completely of other things. Like credits consume all channels. Perhaps the amount of channels was different by default before, I'm investigating a little bit here.

mvanhorn:
Glad to get this in, thanks @stefanhendriks. The music volume only responding to the Music slider now is the right behavior.

mvanhorn:
Appreciate getting this in, @stefanhendriks. Good to have the Sound slider stop yanking the music volume around.

PR #1409, opened 2026-06-19
Guard against dead-unit dereference in saboteur self-destruct path

### LINKED ISSUE #1404: Crash when using Saboteur to destroy a rocket turret
Not sure what caused this, but i'm guessing it has something to do with rocket turrets as the game crashed when a used a Saboteur to destroy one

[nlog.zip](https://github.com/user-attachments/files/29009104/nlog.zip)

mvanhorn:
## Summary

- Guards `onMyStructureAttacked` in `cPlayerBrainCampaign` against accessing an attacker unit that died before the event was processed
- Guards `cUnit::die()` against being called a second time on a slot already reinitialized in the same tick — `bRemoveMe` is set first so `isValid()` reliably returns false only for re-initialized slots (see #1404)
- Includes `eventType` in all `s_GameEvent::toString()` variants for clearer debug output
- Event emission logging gated behind `-debug` mode; unit init logging downgraded to DEBUG level
- Guards `cAbstractStructure::die()` against calling `init()` on a unit already mid-execution in its own `die()` (defensive, for the `iUnitIDWithinStructure` path)
- Skips explosion damage in `createExplosionParticle()` for units already marked for removal — fixes the confirmed crash path

## Root causes

**Crash 1 — double-die:** When multiple saboteurs attack the same structure simultaneously, a unit slot can be killed and reinitialized in the same thinkFast pass. A second call to `die()` on that slot tries to build a `GAME_EVENT_DESTROYED` event using `getPlayer()`, which hits `cPlayers::getPlayer(-1)` and asserts. Making `die()` idempotent via `!isValid()` (with `bRemoveMe = true` set first) fixes this.

**Crash 2 — chained explosion causes recursive die():** A Saboteur or Devastator calls `die()` without first taking damage, so `iHitPoints` stays positive during `die()` execution. The unit is still registered in `MAPID_UNITS` until `remove_id` at the very end of `die()`. If the explosion hits a nearby Devastator, that Devastator's own explosion can re-hit the original dying unit's cell, pass the `iHitPoints > 0` check, and call `die()` recursively. The recursive `die()` passes `!isValid()` (because `bRemoveMe == true` suppresses the HP check) and runs to completion including `init()`, resetting `iPlayer = -1`. The outer `die()` then crashes at `getPlayer()`.

The fix is a `bRemoveMe` guard in the explosion unit-damage loop: units already in the process of dying are skipped entirely, preventing the recursive `die()`.

## Test plan

- Send multiple saboteurs simultaneously at a single structure (the repro from #1404)
- Send a Devastator into a cluster of structures to trigger chained explosions
- Confirm no crash and explosion animations play normally
- Test normal unit combat deaths for regressions

### DIFF
--- a/src/game/cGame_logic.cpp
+++ b/src/game/cGame_logic.cpp
@@ -1169,6 +1169,9 @@ void cGame::setPlayerToInteractFor(cPlayer *pPlayer)
 void cGame::emitGameEvent(const s_GameEvent &event)
 {
     if (m_eventEmitter) {
+        if (m_gameSettings->isDebugMode()) {
+            Logger::trace(eLogComponent::COMP_NONE, "cGame::emitGameEvent", "event {}", s_GameEvent::toString(event));
+        }
         m_eventEmitter->emit(event);
     }
 }
--- a/src/gameobjects/players/brains/cPlayerBrainCampaign.cpp
+++ b/src/gameobjects/players/brains/cPlayerBrainCampaign.cpp
@@ -273,6 +273,11 @@ void cPlayerBrainCampaign::onMyStructureAttacked(const DamagedEvent &event)
     if (unitIdThatAttacks > -1) {
         // respond to something that attacks us
         cUnit *originUnit = m_objects->getUnit(unitIdThatAttacks);
+        if (!originUnit->isValid()) {
+            // attacker (e.g. saboteur) may have self-destructed by the time the damage event is processed
+            log(std::format("Unit {} that attacked my structure is no longer valid, ignoring.", unitIdThatAttacks).c_str());
+            return;
+        }
         if (originUnit->getPlayer()->isSameTeamAs(player)) {
             // friendly fire, ignore
             log(std::format("Unit {} who damaged my structure is from friendly player, ignoring.", unitIdThatAttacks).c_str());
--- a/src/gameobjects/players/brains/cPlayerBrainCampaign.cpp
+++ b/src/gameobjects/players/brains/cPlayerBrainCampaign.cpp
@@ -273,11 +273,6 @@ void cPlayerBrainCampaign::onMyStructureAttacked(const DamagedEvent &event)
     if (unitIdThatAttacks > -1) {
         // respond to something that attacks us
         cUnit *originUnit = m_objects->getUnit(unitIdThatAttacks);
-        if (!originUnit->isValid()) {
-            // attacker (e.g. saboteur) may have self-destructed by the time the damage event is processed
-            log(std::format("Unit {} that attacked my structure is no longer valid, ignoring.", unitIdThatAttacks).c_str());
-            return;
-        }
         if (originUnit->getPlayer()->isSameTeamAs(player)) {
             // friendly fire, ignore
             log(std::format("Unit {} who damaged my structure is from friendly player, ignoring.", unitIdThatAttacks).c_str());
--- a/src/gameobjects/players/cPlayers.cpp
+++ b/src/gameobjects/players/cPlayers.cpp
@@ -67,7 +67,7 @@ const cPlayer* cPlayers::operator[](int index) const {
 }
 
 cPlayer* cPlayers::getPlayer(int index) {
-    d2tm_assert(index >= 0 && index < MAX_PLAYERS_CAPACITY && "cPlayers::getPlayer out of bounds");
+    d2tm_assert(index >= 0 && index < MAX_PLAYERS_CAPACITY && std::format("cPlayers::getPlayer out of bounds: {}\n\n", index).c_str());
     if (index < 0 || index >= MAX_PLAYERS_CAPACITY) {
         return nullptr;
     }
--- a/src/gameobjects/structures/cAbstractStructure.cpp
+++ b/src/gameobjects/structures/cAbstractStructure.cpp
@@ -193,7 +193,14 @@ void cAbstractStructure::die()
 
     // UnitID > -1, means the unit inside will die too
     if (iUnitIDWithinStructure > -1) {
-        m_objects->getUnit(iUnitIDWithinStructure)->init(iUnitIDWithinStructure); // die here... softly
+        cUnit *pUnit = m_objects->getUnit(iUnitIDWithinStructure);
+        if (!pUnit->isMarkedForRemoval()) {
+            pUnit->init(iUnitIDWithinStructure); // die here... softly
+        }
+        else {
+            Logger::debug(COMP_STRUCTURES, "cAbstractStructure::die",
+                "Unit [{}] inside structure [{}] already marked for removal, skipping init.", iUnitIDWithinStructure, id);
+        }
     }
 
     if (iUnitIDEnteringStructure > -1) {
--- a/src/gameobjects/units/cUnit.cpp
+++ b/src/gameobjects/units/cUnit.cpp
@@ -186,6 +186,17 @@ void cUnit::setBoundParticleId(int particleId)
 
 void cUnit::die(bool bBlowUp, bool bSquish)
 {
+    // Guard against die() running on a unit that has already been destroyed and re-initialized
+    // this tick. die() ends by calling init(iID), which resets iPlayer to -1. If a unit destroys
+    // something and then dies within the same thinkFast pass, die() can be invoked a second time
+    // on the now re-initialized unit; the GAME_EVENT_DESTROYED event below calls getPlayer() ->
+    // cPlayers::getPlayer(-1), which asserts (see #1404). iPlayer < 0 only after init(), so it is a
+    // reliable "already dead" sentinel. isValid() cannot be used here: a legitimate first death has
+    // iHitPoints < 0 with bRemoveMe still false, which already makes isValid() return false.
+    if (iPlayer < 0) {
+        return;
+    }
+
     // DO NOTE: We do *not* set the HP to -1 here for a reason. Being: that the isValid() function checks for
     // health and that will give us a unit ID that is the *same* as this unit ID. (see UNIT_NEW() implementation).
     // hence we don't want to fiddle with that (for now), but instead we set a "removeMe" flag.
--- a/src/gameobjects/units/cUnit.cpp
+++ b/src/gameobjects/units/cUnit.cpp
@@ -259,6 +259,11 @@ void cUnit::die(bool bBlowUp, bool bSquish)
         }
     }
 
+    if (iPlayer < 0) {
+        Logger::info(COMP_UNITS, "die()", "iPlayer became < 0 while it was >= 0 just a moment ago!");
+        return;
+    }
+
     // before re-initing, send out event, so in case we need to handle the event and fetch the data from that
     // entity then we can atleast pry it for data...
     const s_GameEvent event {
@@ -397,13 +402,17 @@ void cUnit::createExplosionParticle()
                 int idOfUnitAtCell = m_map->getCellIdUnitLayer(cll);
                 if (idOfUnitAtCell > -1) {
                     int id = idOfUnitAtCell;
+                    cUnit *pHitUnit = m_objects->getUnit(id);
 
-                    if (m_objects->getUnit(id)->iHitPoints > 0) {
-                        m_objects->getUnit(id)->iHitPoints -= 150;
+                    // Skip units already dying: a saboteur/devastator dies with positive HP (die() is
+                    // called before taking damage), so the HP check alone does not guard against a
+                    // recursive die() being triggered by a chained explosion (see #1404).
+                    if (!pHitUnit->bRemoveMe && pHitUnit->iHitPoints > 0) {
+                        pHitUnit->iHitPoints -= 150;
 
                         // NO HP LEFT, DIE
-                        if (m_objects->getUnit(id)->iHitPoints <= 1)
-                            m_objects->getUnit(id)->die(true, false);
+                        if (pHitUnit->iHitPoints <= 1)
+                            pHitUnit->die(true, false);
                     } // only die when the unit is going to die
                 }
 
--- a/src/gameobjects/units/cUnit.cpp
+++ b/src/gameobjects/units/cUnit.cpp
@@ -408,11 +408,13 @@ void cUnit::createExplosionParticle()
                     // called before taking damage), so the HP check alone does not guard against a
                     // recursive die() being triggered by a chained explosion (see #1404).
                     if (!pHitUnit->bRemoveMe && pHitUnit->iHitPoints > 0) {
-                        pHitUnit->iHitPoints -= 150;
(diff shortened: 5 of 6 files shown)


Ripps1:
Love this guy ^

stefanhendriks:
I have been looking into this as well lately. But I am afraid this won't cover it.

A `Saboteur` will have flag `canAttackOnEnterStructure` set to `true`. This means it will go into the code in `cUnit` ~ line 2876:

```cpp
  if (canAttackOnEnterStructure()) {
                            pStructure->damage(getUnitInfo().damageOnEnterStructure, -1); // no need to pass ID of unit, as it is dead
                            die(true, false);
                        }
                        else {
                            if (pStructure->getHitPoints() < pStructure->getCaptureHP()) {
                                // make structure switch sides
                                pStructure->getsCapturedBy(getPlayer());
                                die(false, false);
                            }
                            else {
                                // the unit will die and inflict damage
                                die(true, false);
                                pStructure->damage(getUnitInfo().damageOnEnterStructure, -1); // no need to pass ID of unit, as it is dead
                            }
                        }
```
Meaning, it will first damage structure, which fires a `DAMAGED` event, which is captured by any brain. At that moment in time, the unit itself did not die yet. (as that is the next call `die(true, false)`).

image


Also, this PR focusses only on Campaign player, not Skirmish brain. So the fix would have to be at both places (if it was the actual fix).

I tested this PR on my machine and could not get it to crash. And I am testing Skirmish mode (which does not have any changes, so I would have expected crashes).

I tried to destroy multiple buildings using saboteurs. Turrets, factories, refineries, etc. Nothing really did crash. One crash should help identify it, with a stack trace that is.

@Ripps1 perhaps you have some ideas about in what context the crash occured? It is hard to solve it this way.

stefanhendriks:
Well what do you know, I could cause a crash and I am still not wiser (yet).

image

the `getPlayer()` causes a crash, because `iPlayer` is `-1`, which will result in an out of bounds. 

The weird thing is, `iPlayer` is only reset a few lines later in `init()`. There are not many lines that change the `iPlayer`.

`bRemoveMe` is `false` as well. 

This implies `init` has been called before the die() here. 

However, nothing to be seen here in the stacktrace:

image


One detail. I could cause the crash when I sent a bunch of Saboteurs at once to a building. It might be related by having one saboteur entering structure. Destroying the structure, then destroying itself (causing explosion around it), which can explode additional neighbouring saboteur and so on. Perhaps something happens there in the order of things. 

I'll keep digging.

stefanhendriks:
Thread dump:

```
Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	    0x7ff8198cb7de __pthread_kill + 10
1   libsystem_pthread.dylib       	    0x7ff819908a5b pthread_kill + 259
2   libsystem_c.dylib             	    0x7ff819818494 __abort + 145
3   libsystem_c.dylib             	    0x7ff819818403 abort + 141
4   d2tm                          	       0x1073ab470 d2tm_assert(bool, std::__1::source_location) + 624 (cAssert.h:53)
5   d2tm                          	       0x10742b292 cPlayers::getPlayer(int) + 34 (cPlayers.cpp:70)
6   d2tm                          	       0x10743ab9e cUnit::getPlayer() + 8 (cUnit.cpp:3210) [inlined]
7   d2tm                          	       0x10743ab9e cUnit::die(bool, bool) + 702 (cUnit.cpp:258)
8   d2tm                          	       0x107447b5a cUnit::thinkFast() + 90 (cUnit.cpp:3623)
9   d2tm                          	       0x10745b828 cGamePlaying::thinkFast() + 440 (cGamePlaying.cpp:133)
10  d2tm                          	       0x1073e34db cGame::thinkFast() + 59 (cGame_logic.cpp:1157)
11  d2tm                          	       0x1073ee988 cTimeManager::handleTimerFast() + 25 (cTimeManager.cpp:141) [inlined]
12  d2tm                          	       0x1073ee988 cTimeManager::processTime() + 248 (cTimeManager.cpp:203)
13  d2tm                          	       0x1073e0636 cGame::run() + 342 (cGame_logic.cpp:467)
14  d2tm                          	       0x10747e2c0 main + 352 (main.cpp:57)
15  dyld                          	    0x7ff819531b28 start + 3240
```

stefanhendriks:
I could make it crash again. I have positioned all Saboteurs nearby a structure (any) and order them to attack it at once. 

I could not *always* make it crash though.

Aside from that i added a log statement to see if `init` was called on any unit. This is what the last lines in the log say:

```
249194|INFO|UNITS|Initialization|Initializing unit 4||
249194|DEBUG|SOUND|cGame::playSoundWithDistance|iDistance [179], distanceNormalized [0.89040816] maxDistance [1633.3334], m_zoomLevel [0.9], volumeFactor [0.63], volume [113.972244], iVolFactored [71]|||2
249194|INFO|STRUCTURES|cAbstractStructure::damage|Structure [10] received [206] damage from originId [-1], HP is now [121]||
```

As you can see unit `4` is initialised, this means `iPlayer` is set to `-1`. But, at the same time, this unit is now causing the crash. This makes sense.

I still lack more context.

stefanhendriks:
Added some logging about emitting events. Caused another crash:

```
156995|TRACE|NONE|cGame::emitGameEvent|event CommonEvent[type=GAME_EVENT_DESTROYED, entityType=UNIT, entityId=21, player=present, entitySpecificType=19, atCell=-1, isReinforce=false]||
<snip>
156995|INFO|UNITS|Initialization|Initializing unit 21||
156995|INFO|STRUCTURES|cAbstractStructure::damage|Structure [33] received [208] damage from originId [-1], HP is now [292]||
156995|TRACE|NONE|cGame::emitGameEvent|event DamagedEvent[type=GAME_EVENT_DAMAGED, entityType=STRUCTURE, entityId=33, player=present, entitySpecificType=1, atCell=-1, originId=-1, originType=UNIT]||
<snip>
156995|DEBUG|SOUND|cGame::playSoundWithDistance|iDistance [209], distanceNormalized [0.85782313] maxDistance [1470], m_zoomLevel [1], volumeFactor [0.7], volume [109.80136], iVolFactored [76]|||3
156995|DEBUG|SOUND|cGame::playSoundWithDistance|iDistance [177], distanceNormalized [0.8795918] maxDistance [1470], m_zoomLevel [1], volumeFactor [0.7], volume [112.58775], iVolFactored [78]|||4
156996|TRACE|NONE|cGame::emitGameEvent|event CommonEvent[type=GAME_EVENT_DESTROYED, entityType=UNIT, entityId=16, player=present, entitySpecificType=19, atCell=-1, isReinforce=false]||
<snip>
<crash>
```

You can see, unit `21` is initialised. And this is the unit that caused a crash.

But, if you look up, you can see:

```
156995|TRACE|NONE|cGame::emitGameEvent|event CommonEvent[type=GAME_EVENT_DESTROYED, entityType=UNIT, entityId=21, player=present, entitySpecificType=19, atCell=-1, isReinforce=false]||
```

So, entity 21 got destroyed (a unit). And somehow ran the loop about `thinkFast`, which determined it destroyed something and from there on it calles `die` on itself, but that is invalid. 

The `thinkFast` methods have an `isValid()` guard. Given that it has this guard, it means the code has to be past that point and still seems to execute logic. Meaning, at the moment the unit `thinkFast` was executing, it somehow was not `invalid` yet, but became invalid. 


Trace:
```
Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	    0x7ff8198cb7de __pthread_kill + 10
1   libsystem_pthread.dylib       	    0x7ff819908a5b pthread_kill + 259
2   libsystem_c.dylib             	    0x7ff8198183f4 abort + 126
3   d2tm                          	       0x10a773470 d2tm_assert(bool, std::__1::source_location) + 624 (cAssert.h:53)
4   d2tm                          	       0x10a7f35c2 cPlayers::getPlayer(int) + 34 (cPlayers.cpp:70)
5   d2tm                          	       0x10a802f0e cUnit::getPlayer() + 8 (cUnit.cpp:3211) [inlined]
6   d2tm                          	       0x10a802f0e cUnit::die(bool, bool) + 702 (cUnit.cpp:259)
7   d2tm                          	       0x10a80fbda cUnit::thinkFast() + 90 (cUnit.cpp:3624)
8   d2tm                          	       0x10a8238a8 cGamePlaying::thinkFast() + 440 (cGamePlaying.cpp:133)
9   d2tm                          	       0x10a7ab4db cGame::thinkFast() + 59 (cGame_logic.cpp:1157)
10  d2tm                          	       0x10a7b6cb8 cTimeManager::handleTimerFast() + 25 (cTimeManager.cpp:141) [inlined]
11  d2tm                          	       0x10a7b6cb8 cTimeManager::processTime() + 248 (cTimeManager.cpp:203)
12  d2tm                          	       0x10a7a8636 cGame::run() + 342 (cGame_logic.cpp:467)
13  d2tm                          	       0x10a846490 main + 352 (main.cpp:57)
14  dyld                          	    0x7ff819531b28 start + 3240
```

stefanhendriks:
I need to take a break, I'll ask Claude to investigate with the information gathered so far.

mvanhorn:
Thanks for the detailed trace and logs, that pinned it down.

Root cause: `die()` is not idempotent. The crash is the `GAME_EVENT_DESTROYED` event `die()` emits near the end, where `.player = getPlayer()` resolves to `m_objects->getPlayer(iPlayer)`. `die()` finishes by calling `init(iID)`, which resets `iPlayer` to `-1` (and `bRemoveMe` to false). So once a unit has died, a second `die()` on that same unit builds the destroyed-event with `getPlayer()` -> `cPlayers::getPlayer(-1)`, tripping the `index >= 0 && index < MAX_PLAYERS_CAPACITY` assert at `cPlayers.cpp:70`. That matches your stack exactly: `thinkFast -> cUnit::die -> cUnit::getPlayer -> cPlayers::getPlayer -> abort`.

This lines up with your read of the logs: unit 21 emitted `GAME_EVENT_DESTROYED` (first die), then in the same `thinkFast` pass it "destroyed something" and `die()` ran again on the now re-initialized unit. The upstream `isValid()` guard passes because the unit was still valid when `thinkFast` started; it only became invalid mid-pass, after the first `die()` re-init.

Fix (pushed): return early at the top of `die()` when `iPlayer < 0`:

```cpp
if (iPlayer < 0) {
    return;
}
```

`iPlayer < 0` is only ever true after `init()`, so it is a reliable "already destroyed and re-initialized" sentinel. I deliberately did not use `isValid()` here: a legitimate first death has `iHitPoints < 0` with `bRemoveMe` still false, which already makes `isValid()` return false, so an `isValid()` guard would swallow normal deaths too.

One caveat: I could not reproduce your exact Saboteur-swarm scenario locally (the build now wants SDL3 + gcc-15, which I don't have set up), so this is reasoned from the trace and code rather than runtime-verified. If you can re-run your "all Saboteurs attack one structure at once" repro against this branch, that would confirm the assert is gone. If it somehow still crashes, the next thing I would look at is whether `thinkFast` should itself bail when a unit goes invalid mid-pass, but the `die()` guard should stop the abort regardless.

stefanhendriks:
*(This comment was written by Claude on Stefan's behalf based on investigation of the crash data)*

---

After digging through the code and crash traces, here is what I found:

**`iUnitIDEnteringStructure` is NOT set for a saboteur.** `startEnteringStructure()` (which sets that field) is only called at `cUnit.cpp:2732`, and that code path is only reachable via `INTENT_UNLOAD_SPICE` or `INTENT_REPAIR`. A saboteur always arrives with `INTENT_CAPTURE` or `INTENT_MOVE`, which goes directly into the `canAttackOnEnterStructure()` branch at line 2876. So when the structure dies, `cAbstractStructure::die()` line 200 does **not** call `die()` on the saboteur via that slot.

**The actual crash is a double-die on the same unit slot.** From the stack traces and logs:

1. A unit's `thinkFast()` passes the `isValid()` guard and starts executing.
2. Somewhere *inside* that execution — as a side effect of the saboteur chain (one saboteur destroys a structure whose explosion reaches a neighbouring saboteur) — `die()` is called on this unit for the first time. That die() emits `GAME_EVENT_DESTROYED` (visible in the log), then calls `init(id)` which resets `iPlayer = -1`.
3. `thinkFast_move()` is still on the call stack and later reaches another `die()` call (line ~2879).
4. That second `die()` tries `getPlayer()` at line 259 with `iPlayer == -1` → assert/crash.

The existing fix (guard in `cPlayerBrainCampaign::onMyStructureAttacked`) does not touch this path. The saboteur passes `-1` as the `originId` in its `pStructure->damage()` call, so `unitIdThatAttacks > -1` is false and the new `isValid()` guard there is never reached.

**The correct fix** is a guard at the top of `cUnit::die()`:

```cpp
void cUnit::die(bool bBlowUp, bool bSquish) {
    if (iPlayer < 0) {
        return;
    }
    ...
```

After `init()` sets `iPlayer = -1`, any subsequent `die()` call on that slot returns immediately. This covers all double-die scenarios without affecting any normal death path (a live unit always has `iPlayer >= 0`).

stefanhendriks:
After applying the fix on this PR, and a clean (re)build, I still get a crash.

stefanhendriks:
Ok, interestingly, I added an additional `iPlayer < 0` check just before creating the event, and it seems to trigger:

```cpp
    if (iPlayer < 0) {
        Logger::info(COMP_UNITS, "die()", "iPlayer became < 0 while it was >= 0 just a moment ago!");
        return;
    }
```

So that means somewhere in the part within `die()` itself, after the initial check, it still breaks. A `die()` call within a `die()` call possibly. Asked Claude to help investigate.

stefanhendriks:
Claude seems to have found something. I'll get back to this shortly.

stefanhendriks:
So far Claude is wrong, the guard I place does stop the crash, but it is not yet pinned down *how* the `iPlayer` becomes `-1` between the guard check at the very beginning in `die()` and the one creating the event.

stefanhendriks:
Ok Claude helped identifying the root cause.

The code responsible for dealing damage (and calling `die` again) is when handling explosions. Explosions are caused within the `die()` loop and can cause a 'double-die' call and mess up state causing the `iPlayer` to become `-1` in between the 1st and 2nd `die()` call.

All very messy for sure, needs refactoring.

For now, there is an if-guard for `bRemoveMe`, if set, it will not handle taking damage and such; effectively avoiding calling `die` within `die`. 

I have tested this with a bunch of saboteurs. It would also have been happening for devastators, btw. 

I can't tell for sure this fixes *all*, but afaik this does cover it. On green build we can merge 🎉

stefanhendriks (approved):

mvanhorn:
Thanks @stefanhendriks - guarding onMyStructureAttacked against a saboteur that already died is a good defensive catch.

PR #1416, opened 2026-07-09
Skirmish maps with fewer than 2 start positions are marked invalid

### LINKED ISSUE #1070: Do you want to join house Generalhouse (none)
Steps to reproduce:

Open 'New map editor'
create a map with no start cells.
save it
open skirmish, start a game on that map.

Image

stefanhendriks:
## Summary

- `cPreviewMaps::loadSkirmish()` now counts valid start cells after loading them. If fewer than 2 are found, `validMap` is set to `false` and a warning is logged. Such maps appear greyed-out and cannot be selected in the skirmish map list.
- Added `cPreviewMaps::isValidMap(int)` that combines the index bounds check and the `validMap` flag. Both the Start button and the Enter key path now use it.
- `D2TM_BUILD_DATETIME` is used directly at the draw call site in `cMainMenuState`; no member variable needed.

Previously a map with no start cells passed the validity check, leading to an out-of-bounds vector access in `prepareSkirmishGameToPlayAndTransitionToCombatState()` and an eventual broken game state (reported as the "Generalhouse (none)" mentat screen).

Closes #1070

## Test plan

- [ ] Create a map in the New Map Editor with no start cells, save it, open Skirmish — the map should appear greyed-out and be unselectable
- [ ] Create a map with exactly 1 start cell — same result
- [ ] Create a map with 2+ start cells — selectable and game starts normally
- [ ] Existing skirmish maps are unaffected

### DIFF
--- a/src/gamestates/cMainMenuState.h
+++ b/src/gamestates/cMainMenuState.h
@@ -6,6 +6,7 @@
 #include "controls/sMouseEvent.h"
 
 #include <memory>
+#include <string>
 
 class Texture;
 class cTextDrawer;
@@ -39,11 +40,10 @@ private:
     int mainMenuWidth;
     int mainMenuHeight;
 
-    cRectangle sdl2power;
-
     std::unique_ptr<GuiWindow> gui_window;
     std::unique_ptr<GuiButton> gui_btn_credits;
 
     Texture *bmp_D2TM_Title;
     Texture *backGroundDebug = nullptr;
+    std::string m_buildStr;
 };
--- a/src/gamestates/cMainMenuState.h
+++ b/src/gamestates/cMainMenuState.h
@@ -6,7 +6,6 @@
 #include "controls/sMouseEvent.h"
 
 #include <memory>
-#include <string>
 
 class Texture;
 class cTextDrawer;
@@ -45,5 +44,4 @@ private:
 
     Texture *bmp_D2TM_Title;
     Texture *backGroundDebug = nullptr;
-    std::string m_buildStr;
 };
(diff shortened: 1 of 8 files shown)


src/gamestates/cMainMenuState.h
```diff
@@ -39,11 +40,10 @@ class cMainMenuState : public cGameState {
...
     int mainMenuHeight;
 
-    cRectangle sdl2power;
-
     std::unique_ptr<GuiWindow> gui_window;
     std::unique_ptr<GuiButton> gui_btn_credits;
 
     Texture *bmp_D2TM_Title;
     Texture *backGroundDebug = nullptr;
+    std::string m_buildStr;
```
stefanhendriks: is it needed to have a string member variable? Can't we simply do this at the line:

```cpp
m_textDrawer->drawTextBottomRight(m_buildStr, 20);
```

src/gamestates/cSetupSkirmishState.cpp
```diff
@@ -277,7 +277,7 @@ cSetupSkirmishState::cSetupSkirmishState(sGameServices* services, cPreviewMaps*
             .withTheme(theme)
             .withKind(GuiRenderKind::OPAQUE_WITH_BORDER)
             .onClick([this]() {
-                if (iSkirmishMap > -1) {
+                if (iSkirmishMap > -1 && m_previewMaps->getMap(iSkirmishMap)->validMap) {
```
stefanhendriks: Perhaps even better is a function like so:


```suggestion
                if (m_previewMaps->isValidMap(iSkirmishMap)) {
```

Where it checks for valid id (> -1) and the map is valid.

stefanhendriks (commented):
2 remarks

Issue #584, opened 2023-01-25
Refactor Carry-all logic so that "bringing new unit" and "transferring existing unit" is the same


stefanhendriks:
Atm the code has a distinction for carry-all's bringing a new unit (it will spawn a new unit when it is at the destination) - or when transferring an existing unit (it remembers the ID of the unit being transferred, the unit that is being transferred is in a 'sort of dead' state, which is used to make it disappear on the map and not interact with other units/bullets/etc).

I think a better way would be to have a unit always spawned, then make it being transferred. Meaning, when a 'new' unit has to be brought to the map (ie a Harvester to a refinery), it should already be spawned. This makes the "how many harvesters do I have" logic a lot easier too when not having to deal with the weird "new to be spawned unit type" edge case.

And, even better, we might be even be able to make a carry-all to bring over more than 1 unit. Simply by having a vector of unit ids to bring over instead of 1 id. (used for OG Dune 2 reinforcements).

stefanhendriks:
PR #1417 (fixing #1302) is a tactical fix that lives inside the "sort of dead" state described here: it makes `isDead()` no longer misreport carryall-carried units, so the player-defeat check stops declaring a loss when all ground units are airborne.

It does not remove the hack. When this refactor lands (real unit always spawned, flagged as being transferred), the `iTempHitPoints` field, the hidden-as-dead handling, and the `isAlive()` helper #1417 adds should all be able to go away.

PR #1417, opened 2026-07-11
isDead() returns false for carryall-carried units

### LINKED ISSUE #1302: Game ends during skirmish game when destroying enemy buildings / units
I was playing a game in skirmish as ordos vs the harkonnen and i was capturing their units and sending them back at them while reinforcing with my own army, i was cheating in large amounts of credits for both sides and was using cheat mode in the earlier portions of the level and it cheat mode was disabled again when i build a palace to avoid sab spam.

Image

After destroying a large portion of their base i was suddenly given the message that i had lost.
what was unusual was technically as far as i could see neither side had technically lost as you can see from the radar image both sides had structures, however it does appear that the AI no longer has any mobile units, this could perhaps be triggering an early defeat condition, but for the wrong house? 

[log.zip](https://github.com/user-attachments/files/28686707/log.zip)

attached is a very large file containing mostly errors about music.

if i have find more information on this will comment here.

stefanhendriks:
## Problem

When a carryall picks up a unit, the code sets `iHitPoints = -1` (a map-hiding hack) and stashes the real HP in `iTempHitPoints`. `isDead()` was `iHitPoints <= 0`, so it returned true for carried units.

`hasAliveUnit()` in the defeat check then skipped those units. If all of a player's ground units were inside carryalls with no structures left, `evaluateStillAlive()` returned false and the player was wrongly declared defeated.

Several call sites in `cPlayer.cpp` already worked around this with `isDead() && !isHidden()` and a *"hidden units play dead"* comment, a sign the predicate was wrong at the source.

## Fix

`isDead()` now requires both HP fields to be negative:

```cpp
// before
bool isDead() { return iHitPoints <= 0; }

// after — carried units keep iTempHitPoints > 0, so they are not dead
bool isDead() { return iHitPoints <= 0 && iTempHitPoints < 0; }
```

New `isAlive()` helper (`isValid() && !isDead() && !isHidden()`) replaces the scattered `!isValid() ... isDead()` guard pairs in the viewport/type getters, the gun turret target checks, and `think_attack` / `think_attack_sandworm`. The `isDead() && !isHidden()` workarounds in `cPlayer.cpp` collapse back to a plain `isDead()`.

## Diagnostic logging

`evaluatePlayerStatus()` now logs when a player is marked defeated, so future reports are actionable without a debug build.

Closes #1302


stefanhendriks:
Related: #584 is the deeper refactor here. This PR fixes the defeat-check bug (#1302) from *inside* the current "sort of dead" carryall hack, by making `isDead()` stop misreporting carried units (`iTempHitPoints < 0` guard) and collapsing the scattered `isDead() && !isHidden()` workarounds into an `isAlive()` helper.

#584 proposes removing the hack entirely (always keep a real spawned unit, flag it as *being transferred*), which would obsolete `iTempHitPoints`, the hidden-as-dead handling, and most of the `isAlive()` helper this PR introduces.

PR #1420, opened 2026-07-15
Deviator ignores orders to attack structures

### LINKED ISSUE #1089: Deviator unit should ignore orders to attack structures since they can do no damage
if a Deviator unit should be given an order to attack a structure it should ignore the order and not move. since they can do no damage.
this means when controlling your army as one group, deviators will not try and get up close and path in front of other more destructive units.

this will further improve the deviator mechanic as now it will be where it belongs, at the back of your army

stefanhendriks:
Fixes #1089

Deviators do no damage to structures. Previously, giving a group attack order against a structure would still send Deviators chasing after it, causing them to path in front of the group's more destructive units instead of staying at the back where they belong.

cUnit::attackStructure() now ignores the order entirely for Deviators, so the unit stays put and doesn't move toward the target.

## Test plan
- [x] ./rebuildmacos.sh builds clean, all 56 unit tests pass
- [x] ./create_release.sh packages successfully

### DIFF
--- a/src/context/cInfoContextCreator.cpp
+++ b/src/context/cInfoContextCreator.cpp
@@ -147,6 +147,7 @@ void cInfoContextCreator::initUnits(cUnitInfos& unitInfos)
         // attack related
         unitInfo.canAttackAirUnits = false;
         unitInfo.canAttackUnits = false;
+        unitInfo.canDamageStructures = true;
 
         // capturing / damage upon entering structure related
         unitInfo.canEnterAndDamageStructure = false;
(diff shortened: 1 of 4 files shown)


stefanhendriks:
Re: the `attackAt()` early-return note — acknowledged, but intentionally leaving it as-is:

- Not a regression. The prior `isType(DEVIATOR)` guard in `attackStructure()` had the identical early `return`; this change only moved the condition to a unit-type property.
- The trigger state does not occur in normal play: a cell's unit layer and structure layer are not both populated (foot units path around structures, they do not stand on a structure tile). The worm layer is handled separately after.
- Adding fallback branching to `attackAt()` for a cell state that cannot arise is scope creep past this change and against the repo's simplicity-first guidance.

Not acting on this. PR is ready to merge.

PR #1421, opened 2026-07-15
Structure placement adjacent to enemy structures blocked

### LINKED ISSUE #989: It should be impossible to place structures adjacent enemy structures
See:

Image

it is possible to simply build a structure, and place it near an enemy structure. The current placing logic only checks if a structure is adjacent, but does not check if it is for the same player.

The fix:
Only allow adjacent to same-player structures.

stefanhendriks:
Fixes #989

The build-distance adjacency check compared the adjacent structure's owner against the HUMAN constant (player id 0) instead of the player who is actually placing the structure. This let AI/other players treat player 0's structures as their own for adjacency purposes, and meant the check wasn't actually tied to "is this my own structure" the way it was intended to be.

Both the placement gate (cMousePlaceState::mayPlaceIt) and its duplicated visual preview logic (cPlaceItDrawer::drawStatusOfStructureAtCell) now compare against the placing player's own id.

## Follow-up: enemy foundation slab

Retesting showed placement adjacent to an enemy structure was still allowed. After the owner check rejects a non-owned structure, execution fell through to the wall/slab check for that same cell. Since every structure sits on a TERRAIN_SLAB foundation, an enemy structure next to the hover still satisfied the build-distance check via its own slab.

Fixed by skipping the wall/slab check for cells occupied by a structure that isn't the placing player's. Player-placed slabs and walls still work as build anchors, so building away from structures via slabs is unchanged.

## Test plan
- [x] ./rebuildmacos.sh builds clean, all 56 unit tests pass
- [x] ./create_release.sh packages successfully
- [x] Manual: cannot place a structure adjacent to an enemy structure
- [x] Manual: can still place adjacent to own structures / own slabs

### DIFF
--- a/.claude/review.md
+++ b/.claude/review.md
@@ -12,6 +12,20 @@ General principles (see `CLAUDE.md` for the full version):
   abstraction, configurability, or error handling for impossible states.
 - Match surrounding style rather than imposing a different one.
 
+## Review output format
+
+Keep the posted review short. A maintainer should read it in under a minute.
+
+- One summary comment. Target under 120 words.
+- Structure: a one-line verdict, then a list of issues found. Nothing else.
+- Each issue: `file:line` — what is wrong — suggested fix. One or two sentences.
+- If nothing needs changing, say so in one line and stop.
+- Do not narrate your process. No "Checked:" / "Verified:" list, no progress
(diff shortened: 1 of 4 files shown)


stefanhendriks:
functional testing shows that this issue is not fixed yet, will continue working on it.

PR #1440, opened 2026-08-30
Bugfix/window letterbox black bars ?

Mira-Chronos:
Try to fix the issue from kilo525
> "Oh I forgot to ask, how to make it so that lower resolutions still display to the full screen? I used to be able to play 1280x720 fullscreen just fine in the previous version, but when I do it now it doesn't scale to my actual monitor so the game is just a rectangle surrounded by black in the middle of my screen"

## **Goal**

Change SDL3 migration format to original SDL2 LETTERBOX  that were converted to INTEGER_SCALE to be able to zoom in fractional factor

### Changes
- restore letterbox behavior

## Testing Steps
- no test, i didn't have the correct OS

### DIFF
--- a/src/game/cGame_logic.cpp
+++ b/src/game/cGame_logic.cpp
@@ -618,7 +618,7 @@ bool cGame::setupGame()
     }
     Logger::info(COMP_INIT, "Initializing Key Bindings", "Loaded from [KEYS] section");
 
-    m_Screen = std::make_unique<cSDLSystem>(m_gameSettings->m_screenW, m_gameSettings->m_screenH, title);
+    m_Screen = std::make_unique<cSDLSystem>(m_gameSettings->m_screenW, m_gameSettings->m_screenH, title, m_windowed);
     if (!m_windowed) {
         m_Screen->setFullScreenMode();
     }
--- a/src/game/cSDLSystem.cpp
+++ b/src/game/cSDLSystem.cpp
@@ -8,16 +8,23 @@
 #include <format>
 #include <stdexcept>
 
+#include <algorithm>
 #include <array>
 #include "include/cAssert.h"
 #include <optional>
 #include <span>
 #include <vector>
 
+// smallest coordinate space the game UI is laid out for
+static constexpr int MIN_RENDER_WIDTH = 800;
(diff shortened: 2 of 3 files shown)


stefanhendriks:
@claude re-review

stefanhendriks:
I can reproduce this on mac as well with the resolution specified running in full-screen. So I will try now this branch and see if the black bars are gone and scaling works as expected.

stefanhendriks:
I am testing this, and this re-introduces the weird scaling issues we got before

image

As you can see the text looks odd.

We could accept this though, as it is an odd resolution and stretches to full-screen on my mac which does not like the dimensions. We could a/b test it, offer this as a release to test for others before we decide.

wdyt @Ripps1 / @Mira-Chronos ?

Mira-Chronos:
As I see it, SDL3 simply does what the user asks of it. Consequently, if it has to force resolutions onto screens not designed for them, the game will end up looking truly awful.

I also consider this normal behavior. Bear in mind that an old 4:3 monitor doesn't need to make any adjustments, so the image quality won't look as blurry as the screenshot you took.

That is also why you can't simply emulate a 640x480 resolution within a 1200x800 render for display on a 1920x1080 screen.

Is the software being too permissive with resolutions? Should we really allow so many options? The software offers a wide range of display settings; it is up to the user to choose the one that suits them best.

stefanhendriks:
@Ripps1 heads up on this one: fullscreen scaling mode is now a configurable feature flag instead of being hardcoded.

While testing this branch, forcing LETTERBOX everywhere turned out to reintroduce the old scaling issues (jagged text at fractional scale), so it defaults back to IntegerScale (the pre-PR behavior) and LETTERBOX is now opt-in for anyone who needs the game to fill the screen edge-to-edge on a resolution that isn't a clean multiple.

You can pick the mode three ways:
- `settings.ini`: `ScalingMode=IntegerScale` or `ScalingMode=Letterbox`
- Command line: `-integerscale` or `-letterbox`
- In-game hotkey: `Alt+L` toggles it live

A manual build with this change is up here for testing: https://github.com/stefanhendriks/Dune-II---The-Maker/releases/tag/manual-b4a9a49

PR #1445, opened 2026-09-11
Sound effect tracks no longer stolen or clicked when stopped

### LINKED ISSUE #1444: Sound effects get cut off when many sounds play in quick succession
Reported by Ripps (via feedback), present since v0.8.5-rc1.

Sound effects don't fully play, especially when many sounds trigger in quick succession. Example: repeated "money tick" sounds (credits counting up) combined with a longer sound effect causes the longer sound to get cut off. Ripps described it as "the max number of sounds at any moment was reduced from a very large number to a very small one."

## Root cause

`cSoundPlayer` (src/utils/cSoundPlayer.cpp) pools 32 `MIX_Track`s for sound effects and picks the next one with a blind round-robin index (`m_nextSfxTrack`), unconditionally calling `MIX_StopTrack` on whatever was previously assigned there, even if it is still playing. If more than 32 sound-triggering events happen before a longer sound naturally finishes, its track gets stolen and stopped prematurely, regardless of whether other tracks in the pool are actually free.

This was introduced in the SDL3_mixer rewrite (replacing the old SDL2_mixer `Mix_PlayChannel(-1, ...)` call, which picks a free channel automatically instead of blindly cycling).

stefanhendriks:
## Summary
- Ripps reported audio getting cut off, present since v0.8.5-rc1: when a lot of sound effects trigger in quick succession (e.g. repeated credit-tick sounds), a longer sound in progress gets cut off. He described it as "the max number of sounds at any moment was reduced from a very large number to a very small one."
- Root cause: `cSoundPlayer` pools sound-effect tracks and picked the next one with a blind round-robin index, unconditionally calling `MIX_StopTrack` on whatever was already assigned there, even if still playing. Introduced in the SDL3_mixer rewrite, replacing the old `Mix_PlayChannel(-1, ...)` call which picked a free channel automatically.
- Fix: extracted the track-selection policy into `pickSfxTrack()` (src/utils/SfxTrackAllocator.cpp), which scans the pool starting at the round-robin position for a track that has actually finished playing, and only steals when every track in the pool is busy.
- Added mixer-level fade-in/fade-out since `cSoundPlayer` was doing zero-frame instant starts/stops, audible as clicks.
- Fixed 10 sound assets on the Data repo side whose raw sample data itself didn't decay to silence at the tail and/or had a hard onset jump, which the mixer-level fade alone can't smooth over for a track reaching its own natural end. `SOUND_PLACE` needed both a tail and an onset fix (the onset was missed by the first scan since it checked sample[0], which was already near zero; the actual jump was at sample[1]).
- Ripps then found voice dialogue still getting cut off under very high game speed with heavy combat/economy sound activity, tracked to `playVoice()` sharing the same pool as every other sound effect with no prioritization. Fixed by doubling the pool (32 -> 64) and making `pickSfxTrack()` priority-aware: voice dialogue now plays at a higher priority and can't be stolen by generic SFX unless every track already outranks it. A more robust fix (a dedicated track reserved for voice) is filed separately as #1446, and the broader architectural gap (no real sound manager, no automated asset validation) is filed as #1447 for 0.9.0.
- Also addressed a review nit: `playSound()` was heap-allocating a fresh `std::vector<bool>` every call on the busy-sound-effects hot path; now reuses member buffers sized once at construction.

Fixes #1444

## Test plan
- [x] `./create_release.sh` packages successfully
- [x] Ripps confirms the original cut-off is fixed
- [x] Ripps confirms most sounds are clearer after the mixer fade fix
- [x] SOUND_PLACE tail and onset fixed at the data level, along with 8 other assets
- [ ] Ripps confirms voice dialogue no longer cuts off under heavy load (priority-aware allocator, pool doubled)
- [ ] Possible residual pop from resampling low-rate legacy assets against the mixer's 48kHz output under separate investigation (candidate -3dB headroom fix, unconfirmed)


stefanhendriks:
@Ripps1 here's a test build with the fix, in case it's handy: https://github.com/stefanhendriks/Dune-II---The-Maker/releases/tag/manual-7caa99c (Windows/Linux/macOS binaries).

Ripps1:
Seems fine now besides the 'popping' noise when placing concrete slabs and in a few other places i cant quite name.
i'll spend the weekend playing the latest builds see what turns up.

stefanhendriks:
@Ripps1 new build with the fade-out fix for the popping noise you mentioned: https://github.com/stefanhendriks/Dune-II---The-Maker/releases/tag/manual-43193a9 (Windows/Linux/macOS binaries). Curious if it makes a difference on your end.

stefanhendriks:
@Ripps1 latest build (internal cleanup only, no audible changes from the previous one): https://github.com/stefanhendriks/Dune-II---The-Maker/releases/tag/manual-5f5c214 (Windows/Linux/macOS binaries). No need to re-test specifically for this one, just keeping you on the latest.

stefanhendriks:
@Ripps1 latest build with a fade-in added (in addition to the fade-out from before): https://github.com/stefanhendriks/Dune-II---The-Maker/releases/tag/manual-0a1e4f8 (Windows/Linux/macOS binaries).

Heads up: we tracked the concrete-slab pop down further, it's actually baked into the SOUND_PLACE sample data itself (very low quality 8-bit/4310Hz legacy sample that doesn't end at silence), not something the mixer engine fix alone can fully solve. Looking into fixing the asset itself next, so that specific sound may still pop a bit in this build, most other sounds should be cleaner now though.

Ripps1:
the asset is still buggy as expected but the rest of the sounds do sound much clearer however i noticed a popping on the build ticker now, when building a radar building

https://github.com/user-attachments/assets/7205c2f4-1272-4879-a7eb-e73f0646f536

Skip to 27 seconds in to hear it

stefanhendriks:
@Ripps1 SOUND_PLACE itself is now fixed too (fixed the actual sample data, it had a hard non-zero ending baked in, not something the mixer engine fix alone could smooth over): https://github.com/stefanhendriks/Dune-II---The-Maker/releases/tag/manual-465df42 (Windows/Linux/macOS binaries).

The radar build sound you flagged has the exact same underlying issue in its sample data, that one's being worked on separately along with a few others we found in a broader scan, will follow up once that's ready.

Ripps1:
So testing Snapshot 465df42 (2026-09-12 08:47) at very high speed (max game speed) rushing through the campaign, i noticed the sounds being culled again, specifically the voice dialogue for when events happen (harvester deployed, construction complete etc.) but it seems to happen mostly when the money ticker is firing and whilst combat is happening bringing back to my earlier suspicion that the max number of sound channels has been dramatically reduced.

Ripps1:
Here is a link to a video showing a short 50 second gameplay session
https://faaf.me/191056d9

Have your AI examine the file since it cant be uploaded to GH directly.

at around 44 seconds to 47 seconds into the video, it can be heard "enemy unit ..." and it should be saying " enemy unit approaching " but its as if the sound is being cut short because the channel it is using to play it has been yanked out from under its feet to be used by another sound who is requesting to play.

stefanhendriks:
@Ripps1 new build, please give this one another go when you can: https://github.com/stefanhendriks/Dune-II---The-Maker/releases/tag/manual-3ce34b8 (Windows/Linux/macOS binaries).

Two things changed since last time:
1. Your "enemy unit approaching" cut-off report at high game speed: found voice dialogue was sharing the same 32-track pool as every other sound effect with no prioritization. Doubled the pool (32 -> 64) and made track-stealing priority-aware, so voice lines can no longer be stolen by regular combat/economy SFX unless every track is itself playing voice dialogue. Would be great if you could specifically retest that high-speed/heavy-combat scenario from your video.
2. SOUND_PLACE's onset click is now fixed too (it had the same defect as the tail, just at the very start of the sound), on top of the tail fix from before.

Filed a couple of follow-ups for later (not blocking this release): #1446 for giving voice its own fully dedicated track, and #1447 for a proper sound manager + automated asset scanning for the 0.9.0 milestone, based on everything we learned chasing this.

PR #1456, opened 2026-09-14
Warnings are only fatal on CI, not local dev builds

### LINKED ISSUE #961: `-Werror` only on release
I'm blocked with flag `-Werror` when i develop 

Fix all warning as error forces me to make larger commits sometimes, which is counterproductive.

I'm not against this flag, but it's not applied consistently.

Couldn't it be enforced on some actions and not others?

stefanhendriks:
## Summary
Fixes #961. `-Werror` was in `D2TM_COMMON_WARNINGS` and applied unconditionally on every platform branch in `CMakeLists.txt`, regardless of build type. Since local builds default to `CMAKE_BUILD_TYPE=Debug` when unspecified, that meant a local, work-in-progress build could be blocked by a single warning mid-change — exactly what the issue reporter (Mira) was complaining about.

Fix: `-Werror` is now added conditionally:
- Automatically when `CI` is set in the environment. GitHub Actions sets `CI=true` for every job on GitHub-hosted runners, and this repo's three workflows (`build_pr.yml`, `build_master.yml`, `build_manual.yml`) all run exclusively on GitHub-hosted runners (`ubuntu-latest`, `windows-latest`, `macos-15`, `macos-15-intel`) — so **no workflow file changes were needed**, CI keeps enforcing warnings-as-errors exactly as before.
- Or explicitly via `-DD2TM_WERROR=ON`, so warnings-as-errors can still be reproduced locally on demand (e.g. to double-check a change before pushing). Documented in the README.

Local builds without either of those are unaffected by warnings (still shown as warnings, just not fatal).

## Test plan
- [x] Verified locally: `cmake ..` (no `CI`, no flag) -> 0 occurrences of `-Werror` in `compile_commands.json`
- [x] Verified locally: `CI=true cmake ..` -> `-Werror` present
- [x] Verified locally: `cmake .. -DD2TM_WERROR=ON` -> `-Werror` present


stefanhendriks:
@Mira-Chronos since you created the ticket in GH, is the solution provided here sufficient for you?

PR #1458, opened 2026-09-14
All src files get the standard MIT license header

### LINKED ISSUE #1067: Licence in code
Why not apply this script with modified line 4 to write in all `src`  the file cartridge for the property ?

[copyright.py](https://github.com/user-attachments/files/27299946/copyright.py)

### LINKED ISSUE #1388: Rewrite all head files with right licence
All in title

We have a script that do that without manual intervention (for 97%  ...)

stefanhendriks:
## Summary
Fixes #1067, fixes #1388. Standardizes the license header across every file in `src/`, using the format Stefan proposed directly on #1067.

Before this: 28 files carried an old "Dune II - The Maker / code by Stefan Hendriks" header with inconsistent year ranges, and the other 353 files had no header at all.

After this: all 381 `.h`/`.hpp`/`.cpp` files under `src/` start with the same MIT-license header (matching `LICENSE.md`), each with its own `@file` name.

Done via a new script, `scripts/license_headers.py` (per Stefan's suggestion in #1067 to keep a script in the repo for this), so it's re-runnable idempotently for any future file that's missing the header, rather than a one-off manual edit. It:
- Skips any file that already has the new header.
- Strips the old-style header block where present (including the one file, `d2tm_math.cpp`, that had extra "Purpose of this file" text folded into the same comment block).
- Prepends the new header to everything else.

`tests/` was intentionally left out of scope, this only covers `src/` as discussed in the issue.

## Test plan
- [x] Verified 0 files still contain the old header text (`grep -rl "code by Stefan Hendriks" src`)
- [x] Verified all 381 source files now contain exactly one `@file` header
- [x] Spot-checked several files (previously headerless, previously old-style, and the one with extra body text) for correct output


stefanhendriks:
@Mira-Chronos do you agree with this approach?

